PR #40, opened 2023-09-06
Feature/ci cd pipeline

mdr223:
This PR contains an initial draft of the work to automate our deployment and operation of the **dev services**, using Github Actions and Docker Compose.  Similar work for our production services will follow in a separate PR.

NOTES:
- Added `.github/` and `deploy/` directories to project
    - `.github/` contains workflow to be used by Github Actions for CI
    - `deploy/` contains scripts, configuration, and Dockerfiles for deploying each service through CI using `docker compose`
- Created `pyproject.toml` and reorganized directory to follow PEP 621 specification
    - In order to make the project pip installable I needed to reorganize a bit such that it aligned w/PEP 621
    - Specifically, I moved all the top-level code directories under an `A2rchi/` directory
    - I also made changes to imports in an attempt to eliminate all relative imports
- Made small changes to `DataManager` while I was looking at it:
    - Swapped `os.mkdir` —> `os.makedirs` with `exist_ok=True`
        - Prevents error(s) if directory already exists
    - Using os.path.join to reduce chance of configuration error due to missing ‘/’
- Made other minor tweaks for syntax / style
    - Will do a more robust pass through code with `refactor` branches to propose more style changes


mdr223:
NOTE: tomorrow I will resolve the remaining merge conflicts. I will also do a final deployment to test that things are working properly -- at the moment I cannot debug the final two services (mailbox and scraper) b/c someone is heavily using the subMIT machine I am testing on.

A2rchi/chains/chain.py
```diff
@@ -29,16 +28,18 @@ def __init__(self):
...
+            )
         else:
             self.client = chromadb.PersistentClient(path = self.global_config["LOCAL_VSTORE_PATH"])
 
         self.collection_name = self.utils_config["data_manager"]["collection_name"] + "_with_" + embedding_name
         self.vectorstore = Chroma(client = self.client, collection_name = self.collection_name,
                                   embedding_function = self.embedding_model)
 
-        print(" N Colls: ",self.client.get_collection(self.collection_name))
+        print(" N Collection: ",self.client.get_collection(self.collection_name))
```
julius-heitkoetter, line 42: Not sure what "N Collection" refers to here. This print statement outputs the name and some meta info about the collection we are using, as follows:

`name='dev_collection_with_OpenAIEmbeddings' id=UUID('bace30e9-e225-4aed-b70c-63844981ce91') metadata=None`


So maybe we should change the string at the beginning of the print statement to be something like "INFO - collection used: "

A2rchi/chains/chain.py
```diff
@@ -52,20 +53,32 @@ def __init__(self):
...
-                                      embedding_function = self.embedding_model)
-            self.chain = BaseChain.from_llm(self.llm, self.vectorstore.as_retriever(), return_source_documents = True)
-            print(" N Coll: ",self.client.get_collection(self.collection_name).count())
+            self.vectorstore = Chroma(
+                client=self.client,
+                collection_name=self.collection_name,
+                embedding_function=self.embedding_model,
+            )
+            self.chain = BaseChain.from_llm(self.llm, self.vectorstore.as_retriever(), return_source_documents=True)
+            print(f" N Collection: {self.client.get_collection(self.collection_name).count()}")
```
julius-heitkoetter, line 76: This prints out the number of collumns (or entries) in the collection, not the Number of collections as the print statement suggests. We should probably change it to "N entries", "N chunks", or something like it.

A2rchi/utils/mailbox.py
```diff
@@ -1,9 +1,14 @@
...
-import getpass, imaplib, email
+from A2rchi.utils.config_loader import Config_Loader
+from A2rchi.utils.env import read_secret
+
+import email
+import imaplib
+import os
+
+# DEFINITIONS
+ISSUE_ID_OFFSET = 9
```
julius-heitkoetter, line 10: What does this do?

pyproject.toml
```diff
@@ -0,0 +1,66 @@
+[project]
+name = "A2rchi"
+version = "0.1.0"
+description = "An AI Augmented Research Chat Intelligence for MIT's subMIT project in the physics department"
```
julius-heitkoetter, line 4: With Christoph pushing for this project to be more and more general, not sure if we want to specifically include subMTI in the description here.

pyproject.toml
```diff
@@ -0,0 +1,66 @@
...
+requires-python = ">=3.7"
+# sorted alphabetically by last name
+authors = [
```
julius-heitkoetter, line 7: Not sure how all this authorship works, do we need to add Tim?

config/config.yaml
```diff
@@ -72,8 +75,8 @@ utils:
     use_HTTP_chromadb_client: True   # recommended: True (use http client for the chromadb vectorstore?)
     # use_HTTP_chromadb_client: False
     vectordb_update_time: 10
-    chromadb_host: "localhost"
-    chromadb_port: 8000
+    chromadb_host: chromadb
```
julius-heitkoetter, line 78: How does this work? I thought that the chromadb host was the IP address of where the chromadb server was running. In the past, I used localhost as an alias to 127.0.0.1. 

It could be correct (no need to necessarily change it), I just wanna make sure I understand it.

config/config.yaml
```diff
@@ -1,7 +1,7 @@
 global:
-  DATA_PATH: "./data/"
-  ACCOUNTS_PATH: "./.accounts/"
-  LOCAL_VSTORE_PATH: "./data/vstore/"
+  DATA_PATH: "/root/data/"
```
julius-heitkoetter, line 2: Is there a better way to have the folders set? I'm worried on a blank install, it's not obvious that the person installing A2rchi has this `root` directory. However, I know that we do need to make it work for our install on the service account.

.github/workflows/ci-cd.yaml
```diff
@@ -0,0 +1,98 @@
...
+      # boilerplate message and pull repository to CI runner
+      - run: echo "🎉 The job was automatically triggered by a ${{ github.event_name }} event."
+      - uses: actions/checkout@v3
+      - run: echo "The ${{ github.repository }} repository has been cloned to the runner."
+
+      # setup SSH 
+      - name: Setup SSH
+        run: |
+          mkdir -p /home/runner/.ssh/
+          echo "${{ secrets.SSH_PRIVATE_KEY_MDRUSSO }}" > /home/runner/.ssh/id_rsa_submit
```
julius-heitkoetter, line 19: This is just any private key which gets us into the A2rchi service account, correct?

julius-heitkoetter (commented):
Mostly small nit-picky comments. Nothing much needs to be changed, just want to make sure we are on the same page about a few things. 

Overall, also, I wanted to double check that the conversation data is saved reliably when we deploy. I.e. we never overwrite conversation data and it won't be deleted if we remake the repo, etc.

A2rchi/chains/chain.py
mdr223: Ahh okay, I was a bit confused about the print statement as well and tried making it more verbose, but your suggestion looks much better to me.

mdr223: I'll also take a second pass through the code in a subsequent PR to swap print statements for `logging.info` statements, which will have that `INFO` prefix by default

A2rchi/utils/mailbox.py
mdr223: Good question! So on line 105 there is a statement that indexes into a string by 9 characters to get the issue id from a description. One general guideline for code cleanliness is to try to limit the number of integers floating around in the source code, especially when their semantic meaning may be hard to interpret. The common solution is to define these constants with a semantically meaningful name at the top of the file. 

Note that `0`, `-1`, and `1` are notable exceptions to this rule because they are commonly understood to be indexing the beginning and end of lists or correcting off-by-one issues. Also, in application specific scenarios it may be obvious why an integer is chosen, which eliminates the need for defining a constant (e.g. using `n % 2` in an `if` statement to filter for even numbers).

pyproject.toml
mdr223: Sounds good, I can reword this!

pyproject.toml
mdr223: Sure thing I'll add Tim -- but to answer your general question I don't think there's a hard rule about this. I was going based off of the set of people who significantly touched the codebase, but your right that we should include my advisor 😂.

config/config.yaml
mdr223: Great question! This is a Docker (Compose)-specific setting, and when we support multiple deployment scenarios, we'll need to have separate configs to handle this.

In Docker Compose, when a set of containers/services are defined in a `compose.yaml` file, each service can access ports on each other service by using the name(s) assigned in the `compose.yaml`. In our case, I named the Chroma service `chromadb` in our `compose.yaml`, so the application code running inside e.g. the mailbox, data manager, etc. accesses ChromaDB on `chromadb:PORT`.

config/config.yaml
mdr223: So this `/root` directory is guaranteed to exist because the Dockerfiles used by Docker compose are built on the `python:3.10` image, which has this path. If we eventually support a BYOC ("bring your own container") deployment then we will need to change this, but that should happen in its own PR so I think this is fine for now.

.github/workflows/ci-cd.yaml
mdr223: Yes correct, so in theory I can replace this with a different key that works w/A2rchi.

mdr223:
> Mostly small nit-picky comments. Nothing much needs to be changed, just want to make sure we are on the same page about a few things.
> 
> Overall, also, I wanted to double check that the conversation data is saved reliably when we deploy. I.e. we never overwrite conversation data and it won't be deleted if we remake the repo, etc.

To this point, the conversations are written relative to the configured `DATA_PATH`, which right now means they're written to `/root/data/conversations.json`. Our deployment creates a Docker volume the first time we deploy (and re-uses it for subsequent deployments), so conversations will live across deployments and will not be deleted unless we explicitly delete the Docker volume.

A2rchi/utils/mailbox.py
julius-heitkoetter: This makes sense, we should document what you wrote above as a comment for this offset, to make it clear what it is. We can save this for a later PR though

julius-heitkoetter (approved):
Looks great! Thanks for answering the comments and making those few small changes. It all seems clear now. I'll go ahead and give the approval, but it seems to be failing the deployment. Could you look into this? I'm not sure why it is happening. Once this is resolved, we can merge.

PR #45, opened 2023-09-18
changing environment variable to correct name

julius-heitkoetter:
The huggingface token should be under HUGGING_FACE_HUB_TOKEN, not HF_TOKEN for huggingface authentication. Changes are made to reflect this.


julius-heitkoetter:
This PR is not complete. I missed a couple changes. Will tear this down and add a new one.

PR #46, opened 2023-09-18
Dev llama patch

julius-heitkoetter:
The huggingface token should be under HUGGING_FACE_HUB_TOKEN, not HF_TOKEN for huggingface authentication. Also, this should get loaded in as an environment variable in 4 of the services. Changes are made to reflect this.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,6 +15,7 @@ dependencies = [
     "accelerate==0.22.0",
     "backoff==2.2.1",
     "beautifulsoup4==4.12.2",
+    "bitsandbytes==0.41.1",
     "chromadb==0.4.8",
     "clickhouse-connect==0.6.6",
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -15,6 +15,7 @@ dependencies = [
     "accelerate==0.22.0",
     "backoff==2.2.1",
     "beautifulsoup4==4.12.2",
+    "bitsandbytes==0.41.1",
```
mdr223, line 18: is this a needed change or was this something added to produce a commit? either way I'll approve, just curious

mdr223 (approved):

pyproject.toml
julius-heitkoetter: bitsandbytes is needed to quantize  the model.

PR #48, opened 2023-09-20
Release which forces curl to use IPv4

julius-heitkoetter:
Release newest version, including changes to llama2. Most importantly, however, curl is forced to use IPv4 because the IPv6 IP tables are bugging on submit06 (the production deployment machine).


mdr223 (approved):
LGTM

PR #49, opened 2023-09-20
Pdf loader patch ludo

ludomori99:
added pypdf to requirements. small change in loader within data_manager.py in order to fix bug only taking into account first page of document. Also, store page of pdf in vector metadata for further uses.


julius-heitkoetter (approved):
Looks good, thanks for working on this!

PR #54, opened 2023-09-21
hotfix: using our own version of chroma w/out telemetry

mdr223:
Also narrowed scope/lifetime of client objects


mdr223:
bypassing branch protections as this is a hotfix

PR #61, opened 2023-09-25
adding terms/conditions and A2rchi type specification

julius-heitkoetter:
Added a link to terms and conditions (which basically state that we do track data, but it is anonymous and that the service should be used responsibly for the intended purpose) and also a text that on the chat homepage that indicates which version of A2rchi we're using (so that we don't get confused).


mdr223 (approved):
LGTM

PR #77, opened 2023-09-27
Feature/generalize prompts

julius-heitkoetter:
Create ability to specify prompts in the config. This allows A2rchi to be flexible across different applications. 

Additionally, fixed a small bug in commenting logic I noticed in `utils/scraper.py` (the previous code only recognized comments if the "#" was at the very beginning of the line)

### DIFF
--- a/A2rchi/chains/prompts.py
+++ b/A2rchi/chains/prompts.py
@@ -4,8 +4,8 @@ from A2rchi.utils.config_loader import Config_Loader
 
 config = Config_Loader().config["chains"]["prompts"]
 
-def read_prompt(prompt_name, is_condense_prompt=False, is_main_prompt=False):
-    with open(f"config/prompts/{prompt_name}", "r") as f:
+def read_prompt(prompt_filepath, is_condense_prompt=False, is_main_prompt=False):
+    with open(prompt_filepath, "r") as f:
         raw_prompt = f.read()
 
     prompt = ""
--- a/A2rchi/chains/prompts.py
+++ b/A2rchi/chains/prompts.py
@@ -1,27 +1,31 @@
 # flake8: noqa
 from langchain.prompts.prompt import PromptTemplate
+from A2rchi.utils.config_loader import Config_Loader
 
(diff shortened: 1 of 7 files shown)


A2rchi/chains/prompts.py
```diff
@@ -1,27 +1,31 @@
...
 
-condense_history_template = """Given the following conversation between you (the AI named A2rchi), a human user who needs help, and an expert, and a follow up question, rephrase the follow up question to be a standalone question, in its original language.
+config = Config_Loader().config["chains"]["prompts"]
 
-Chat History:
-{chat_history}
-Follow Up Input: {question}
-Standalone question:"""
+def read_prompt(prompt_name, is_condense_prompt=False, is_main_prompt=False):
+    with open(f"config/prompts/{prompt_name}", "r") as f:
```
mdr223, line 8: Minor nit: I would make the argument something like `prompt_filepath` (or `prompt_file`) and then pass that into `open()`, (i.e. `with open(prompt_filepath, 'r') as f:`). Then in the config I would provide the relative (or ideally absolute) path to the file. This way if a future developer reads the config they know exactly where that file is supposed to be in the filesystem, without having to read the code.

config/dev-config.yaml
```diff
@@ -32,6 +32,9 @@ chains:
       - User
       - A2rchi
       - Expert 
+  prompts:
+    CONDENSING_PROMPT: condense.prompt #name of the prompt (in config/prompts) that serves to condense a history and a question into a single question
```
mdr223, line 36: Would just update these to be absolute or relative filepaths (see above)

config/prod-config.yaml
```diff
@@ -32,6 +32,9 @@ chains:
       - User
       - A2rchi
       - Expert 
+  prompts:
+    CONDENSING_PROMPT: condense.prompt #name of the prompt (in config/prompts) that serves to condense a history and a question into a single question
+    MAIN_PROMPT: submit.prompt #name of the main prompt (in config/prompts) which takes in a single question and a context.
```
mdr223, line 37: Same thing as above

mdr223 (approved):
One minor nit, otherwise this all looks good. Will approve now, regardless of whether nit is changed.

PR #80, opened 2023-09-27
moved api call from url /get_chat_response to /api/get_chat_response

julius-heitkoetter:
Simple change, just to keep things clean we should start using the convention that all API calls occur to /api/*

Since we only explicitly do this once, we just have one commit that fixes it.


mdr223 (approved):
Perfect 👍

PR #86, opened 2023-10-04
Feature/add user ratings

julius-heitkoetter:
Completes the front and and back end of the user ratings.

Frontend: 
- there are thumbs up and down symbols which are highlighted when you click on them
- when you click thumbs down, you can add some comments and checkoff some boxes to specify what is wrong
- every `num_responses_until_feedback` (variable in config) response, A2rchi politely asks for feedback (but it is not required). We may want to change this to be a different value than what it is in the config right now. 

Backend:
- Once ratings are given, the front end makes an api call to a "api/like" and an "api/dislike" endpoint. These endpoints store the ratings in the same json file that the conversations are stored. The storing is not necessarily pretty, but since the plan is to soon make a simple database for storing all this, I did not put in work to make it more intuitive. I will describe what is done below. One big thing that technically needs a fix is that the filename "conversations_test.json" is still labeled as "test" and is hardcoded. Again, I didn't fix this because I anticipate it will be overwritten with implementation of the database.

Storage:

- the discussion is stored as a value under the key which is that discussion id. Inside the discussion, there are (up to) two fields: a contents field and a feedback field
- the feedback field's value is a list of feedback dictionaries, which all contain what kind of feedback "like" or "dislike" and which piece of chat content the feedback is pertaining to. "dislike" feedback also has some additional fields for additional contents. 
- the content field's value is a list of tuples which outline the entire discussion. The first element of the tuple dictates who says the value stored in the second element of the tuple. 
- an example can be seen below

image

Lastly, this PR moves the hostname away from the javascript template and into the config.


A2rchi/interfaces/chat_app/app.py
```diff
@@ -197,3 +213,55 @@ def index(self):
...
+        try:
+            # Get the JSON data from the request body
+            data = request.json
+
+            # Extract the HTML content and any other data you need
+            chat_content = data.get('content')
+            discussion_id = data.get('discussion_id')
+
+            feedback = {
+                "chat_content" :  chat_content,
```
mdr223, line 227: Will make a longer comment in PR; but I think we may want to include a `message_id` here, which is unique within a discussion.

A2rchi/interfaces/chat_app/app.py
```diff
@@ -197,3 +213,55 @@ def index(self):
...
+
+            # Extract the HTML content and any other data you need
+            chat_content = data.get('content')
+            discussion_id = data.get('discussion_id')
+            message = data.get('message')
+            incorrect = data.get('incorrect')
+            unhelpful = data.get('unhelpful')
+            inappropriate = data.get('inappropriate')
+
+            feedback = {
```
mdr223, line 252: Will make a longer comment in PR; but I think we may want to include a `message_id` here, which is unique within a discussion.

A2rchi/interfaces/chat_app/static/script.js
```diff
@@ -78,10 +86,87 @@ const getChatResponse = async (incomingChatDiv) => {
...
+
+     // Send an API request with the chat content and discussion ID
+     fetch(API_URL, {
+        method: "POST", // You may need to adjust the HTTP method
+        headers: {
+            "Content-Type": "application/json",
+            "Access-Control-Allow-Origin": "*",
+        },
+        body: JSON.stringify({ 
+            content: chatContent,
```
mdr223, line 116: Will make a longer comment in PR; but I think we may want to include a `message_id` here, which is unique within a discussion.

A2rchi/interfaces/chat_app/static/script.js
```diff
@@ -78,10 +86,87 @@ const getChatResponse = async (incomingChatDiv) => {
...
+        const additionalThoughts = additionalThoughtsInput.value;
+
+        fetch(API_URL, {
+            method: "POST", // You may need to adjust the HTTP method
+            headers: {
+                "Content-Type": "application/json",
+                "Access-Control-Allow-Origin": "*",
+            },
+            body: JSON.stringify({ 
+                content: chatContent,
```
mdr223, line 151: Will make a longer comment in PR; but I think we may want to include a `message_id` here, which is unique within a discussion.

A2rchi/interfaces/chat_app/static/script.js
```diff
@@ -78,10 +86,87 @@ const getChatResponse = async (incomingChatDiv) => {
...
+    const other_image = dislikeBtn.previousElementSibling.querySelector("img");
+    other_image.src = "/static/images/thumbs_up.png";
+
+    const API_URL = "http://t3desk019.mit.edu:7861/api/dislike";
+
+    // Show pop-up form
+    popupForm.style.display = "block";
+
+    //wait for user to submit response
+    submitButton.addEventListener("click", function () {
```
mdr223, line 141: I think we may also want to add this event listener on the "X" (i.e. "close popup") button. This way if a user gives a thumbs down -- but no feedback in the popup -- we still record that they disliked the message.

A2rchi/interfaces/chat_app/static/script.js
```diff
@@ -78,10 +86,87 @@ const getChatResponse = async (incomingChatDiv) => {
...
+        });
+
+        //hide pop up formi
+        popupForm.style.display = "none";
+    })
+}
+
+const closeFeedback = (closeBtn) => {
+    //hide pop up formi
+    popupForm.style.display = "none";
```
mdr223, line 167: Here we should send feedback in case the user didn't click the submit button. The server can choose to ignore this feedback (in favor of richer feedback) if the user actually hit the submit button.

A2rchi/interfaces/chat_app/static/script.js-template
```diff
@@ -4,10 +4,18 @@ const chatContainer = document.querySelector(".chat-container");
 const themeButton = document.querySelector("#theme-btn");
```
mdr223, line 4: See comments I left on `script.js` re: adding `message_id` and submitting dislike feedback on closing of popup window.

A2rchi/interfaces/chat_app/app.py
```diff
@@ -70,7 +70,19 @@ def update_or_add_discussion(data_path, json_file, discussion_id, discussion_con
             print(" ERROR - json_file not found. Creating a new one")
 
         # update or add discussion
-        data[str(discussion_id)] = discussion_contents
+        if str(discussion_id) in data.keys():
```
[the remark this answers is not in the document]
julius-heitkoetter: Fixed!

A2rchi/interfaces/chat_app/app.py
```diff
@@ -70,7 +70,19 @@ def update_or_add_discussion(data_path, json_file, discussion_id, discussion_con
...
+            discussion_dict = data[str(discussion_id)]
+        else:
+            discussion_dict = dict()
+
+        if discussion_contents is not None:
+            print(" INFO - found contents.")
+            discussion_dict["contents"] = discussion_contents
+        if discussion_feedback is not None:
+            print(" INFO - found feedback.")
+            discussion_dict["feedback"] = discussion_dict["feedback"] + [discussion_feedback] if ("feedback" in discussion_dict.keys() and isinstance(discussion_dict["feedback"], List)) else [discussion_feedback]
```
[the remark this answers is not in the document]
julius-heitkoetter: Just in case there is some legacy stuff sitting in the file which isn't a list. We can write the out once we swap over to a database system.

mdr223 (approved):
Changes look good to me!

Issue #63, opened 2023-09-26
Enable Guest Login


julius-heitkoetter:
Allow someone to use the chat service under a "guest" account if they don't want to use the OAUTH login. (cache history)

mdr223:
Closing as this is how users are mainly interacting with A2rchi now, and permissioned login has been put on the backburner for more pressing performance / experiment related issues.

PR #88, opened 2023-10-05
Feature/oauth login

mdr223:
Enabling OAuth login for users with google accounts (i.e. gmail and touchstone). The admin of the A2rchi system can:
1. Choose to use OAuth login or disable it
2. Choose to allow guest login or disable it
3. Green-light a set of users by specifying their emails in configuration. (Only these users may log into the app if guest access is disabled).

Note, I have not made the front-end pretty. I also realize we need to add a logout button on the index.html page.

Right now I'm replicating the app alongside the chat and uploader apps, but the uploader app should ultimately be folded into this app -- at which point we can remove the other two from the repository.

This addresses issue #63


A2rchi/utils/config_loader.py
julius-heitkoetter, line 1: I think this may need to get changed back to what it was? Don't want to have anything stored in our personal folders.

A2rchi/app/db.py
```diff
@@ -0,0 +1,38 @@
...
+    if db is not None:
+        db.close()
+
+def init_db():
+    db = get_db()
+
+    with current_app.open_resource("schema.sql") as f:
+        db.executescript(f.read().decode("utf8"))
+
+@click.command("init-db")
```
julius-heitkoetter, line 29: Just curious what the "@click.command" does?

A2rchi/app/app.py
```diff
@@ -0,0 +1,427 @@
...
+        mit_client = WebApplicationClient(MIT_CLIENT_ID)
+
+    def get_google_provider_cfg():
+        return requests.get(GOOGLE_DISCOVERY_URL).json()
+
+    def get_mit_provider_cfg():
+        return requests.get(MIT_DISCOVERY_URL).json()
+
+# enable CORS:
+CORS(app)
```
julius-heitkoetter, line 226: Eventually will want to move this back into a wrapper class. But for now I think it looks like it works and we can do this later.

julius-heitkoetter (approved):
Just one small fix that needs to happen and everything else looks good. I'll go ahead and approve it, just make sure to look through the comments.

A2rchi/utils/config_loader.py
mdr223: Oh yes, good catch!

A2rchi/app/app.py
mdr223: Moved back into the wrapper

mdr223:
Made some finishing touches to re-architect the code into the FlaskWrapper and deployed from my feature branch into the dev environment. I checked that basic functionality seems to work. Will do one final pass to make sure I don't have any changes that I'm supposed to revert before merging -- but otherwise it should be good to go.

mdr223:
Closing pull request as this has been merged into a staging branch, which will ultimately be merged into main.

PR #90, opened 2023-10-06
Release for 8.01

julius-heitkoetter:
I propose this is the release we want to give to 8.01. It has full feedback enabled between the users and the system, but does not have the new UI and login system. This way, we can take the weekend to implement this system properly, then trigger another release once it is done. 

This PR does not need an approval, but if someone else wants to check over it before we release that would be great.\


config/prod-config.yaml
```diff
@@ -13,9 +13,10 @@ interfaces:
     PORT: 7861
     EXTERNAL_PORT: 7681
     HOST: "0.0.0.0" # either "0.0.0.0" (for public) or "127.0.0.1" (for internal)
-    HOSTNAME: "ppc.mit.edu"  # careful, this is used for the chat service
+    HOSTNAME: "t3desk019.mit.edu"  # careful, this is used for the chat service
     template_folder: "/root/A2rchi/A2rchi/interfaces/chat_app/templates"
     static_folder: "/root/A2rchi/A2rchi/interfaces/chat_app/static"
+    num_responses_until_feedback: 3 #the number of responses given by A2rchi until she asks for feedback.
   uploader_app:
```
julius-heitkoetter, line 20: We may want to increase the num_responsesd_until_feedback to like 5 or 10 so that users don’t get annoyed. Don’t know what the best value is here? People’s thoughts?

mdr223: I think 3 is fine for now, we can increase it if we get negative feedback

mdr223 (approved):
This looks good to me!

PR #98, opened 2023-10-13
Feature/logging

ludomori99:
Add few lines in base.py to somewhat store the prompts. Large room for improvement in details of logging, however not straightforward due to convoluted structure of chaining. it's good to have something working for 801 deployment.

### DIFF
--- a/A2rchi/chains/base.py
+++ b/A2rchi/chains/base.py
@@ -1,5 +1,7 @@
 """Chain for chatting with a vector database."""
 from __future__ import annotations
+from loguru import logger
+from langchain.callbacks import FileCallbackHandler
 
 from A2rchi.chains.prompts import CONDENSE_QUESTION_PROMPT, QA_PROMPT
 from A2rchi.utils.config_loader import Config_Loader
@@ -15,6 +17,7 @@ from typing import Any, Dict, List, Optional, Tuple
 
 # DEFINITIONS
 config = Config_Loader().config["chains"]["base"]
+data_path = Config_Loader().config["global"]["DATA_PATH"]
 
 
 def _get_chat_history(chat_history: List[Tuple[str, str]]) -> str:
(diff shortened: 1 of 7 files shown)


A2rchi/chains/base.py
```diff
@@ -88,19 +91,27 @@ def from_llm(
         combine_docs_chain_kwargs = combine_docs_chain_kwargs or {}
         _prompt = QA_PROMPT
         document_variable_name = "context"
+
+        logfile = data_path+"chain_input_output.log"
```
julius-heitkoetter, line 95: This should probably be done using `os.path.join()`. This way it doesn't matter if `data_path` has a "/" on the end of it or not.

A2rchi/chains/base.py
```diff
@@ -88,19 +91,27 @@ def from_llm(
...
+
+        logger.add(logfile, colorize=True, enqueue=True)
```
julius-heitkoetter, line 97: For documentation purposes, could you add a comment or two above this explaining what this logging is doing specifically.

A2rchi/chains/base.py
julius-heitkoetter: Also, may want to consider "chain_input_output.log" to be a configurable parameter. But I'll leave this to you.

For context, I've been meaning to change "conversations.json" to a configurable parameter.

If you do put it into the config, make sure to write a comment on what it is and any specifications (if it has to end in .log, etc.)

Not required though, just a thought

julius-heitkoetter (changes requested):
Looks good! 2 things that I believe are needed and then one more thing which you may want (or not want) to consider.

A2rchi/chains/base.py
mdr223: Looks good now

mdr223 (approved):
I believe this is ready to be merged 👍

PR #99, opened 2023-10-13
Release root

ludomori99:
release of a2rchi based of ROOT docs


config/prod-root-config.yaml
```diff
@@ -0,0 +1,113 @@
...
+    PORT: 5001
+    HOST: "0.0.0.0" # either "0.0.0.0" (for public) or "127.0.0.1" (for internal)
+    template_folder: "/root/A2rchi/A2rchi/interfaces/uploader_app/templates"
+
+chains:
+  input_lists:
+    - empty.list
+    - miscellanea.list
+    - root-docs.list
+    # - root-forum.list
```
mdr223, line 30: Just leaving a comment / check in case you wanted to uncomment this for release? Just pointing it out in case it was commented out temporarily or something

mdr223 (approved):
Approving as this all looks good, left one comment as a check in case something in the config was meant to be uncommented, but assuming it's intended to be left out this is good to go!

config/prod-root-config.yaml
ludomori99: Meant to be commented. I would like to look more into the HTML scraper before including forum files because despite great potential and some well working examples the behavior seems still a bit unstable depending on the HTML formatting, especially when using HF embeddings. I wanted to merge the basic working branch before we add new functionalities.

PR #100, opened 2023-10-13
Adding meta release for submit06

mdr223:
Adding release for A2rchi-meta, which will run on submit06 and answer support emails. Only requires using mailbox and cleo.


mdr223:
One thing that's definitely still missing / needed is the updated email password details. Once I have those I will update them in the Github Actions workflow and this will be ready to go.

ludomori99 (approved):
Seems right to me, just need the secrets.

Issue #104, opened 2023-10-16
BUG: Malformed JSON is somehow created


mdr223:
Our `conversations_test.json` managed to get into a bad state where for some reason the file either (A) had an invalid end '}' inserted into the file before the real end or (B), had some random text appended onto the end of the file. The error occurs in the file I'm uploading at character 212766 (zero-index).

You can see this by running:
```
$ cat conversations_test.json | cut -c 212700-212800
, "incorrect": false, "unhelpful": true, "inappropriate": false}]}}unt of momentum transferred depend
```

Or by opening the file in a code editor such as VS Code.

For privacy reasons I'm sending the file to you via email instead of creating a Gist.

mdr223:
Believe the issue may be due to a lack of a lock on the like/dislike feedback functions. This also may be consistent with the error in the file occurring in a feedback section. Going to create a hotfix PR for merge into `release-8.01`; we can then merge the commit into `main` (and the other releases) once we've fixed 801.

mdr223:
PR: https://github.com/mit-submit/A2rchi/pull/105

PR #105, opened 2023-10-16
HotFix to Address Issue #104

mdr223:
Please see issue #104 for bug description.

### RCA
I believe the cause of this bug is that the `like` and `dislike` methods of our feedback endpoints do not acquire the `ChatWrapper` lock before writing to the `conversations_test.json` file. This can lead to a race condition where any (two or more) user(s) who submit a conversation and provide feedback simultaneously may cause some inconsistent state in our json file.

### Proposed Fix
We acquire the chat wrapper's lock before writing `like` and `dislike` feedback. We take care to release the lock in the `finally` branch of a `try-except`.

### DIFF
--- a/A2rchi/interfaces/chat_app/app.py
+++ b/A2rchi/interfaces/chat_app/app.py
@@ -212,6 +212,7 @@ class FlaskAppWrapper(object):
         return render_template('terms.html')
     
     def like(self):
+        self.chat.lock.acquire()
         try:
             # Get the JSON data from the request body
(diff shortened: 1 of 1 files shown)


julius-heitkoetter (approved):
Looks great! The “finally” clause exactly what we need here. There will not be a deadlock condition, as there is only one locking variable which is garenteed to be released a short time after it is acquired (either the API response is formatted correctly (in which case we return) or it fails).

PR #106, opened 2023-10-16
Hotfix/malformed json lock race condition

julius-heitkoetter:
Testing for Matt’s PR #105 to address issue #104.


julius-heitkoetter:
Going to override the review required to merge into main, since this is Matt’s changes and I already gave a review with #105

PR #108, opened 2023-10-18
Fix/conversation loggin fix

julius-heitkoetter:
Two small fixes I missed in the logging implementation the first time around:

1. There is now a `meta` section in each discussion for logging. This includes a `time_last_used` which is when the discussion was last used .
2. Initially, the last response from a2rchi was not logged. This is fixed now. 

Tested to see if it was working in action #141

### DIFF
--- a/.github/workflows/dev-ci-cd.yaml
+++ b/.github/workflows/dev-ci-cd.yaml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - fix/conversation_loggin_fix
 jobs:
   deploy-dev-system:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


mdr223 (approved):
This looks good to me!

ludomori99 (changes requested):
Looks great; one note that could improve is to store more than just the timestamp of the last interaction. Would be ideal, as we discussed in the office, to have in metadata a timestamp for when the discussion was created, and all timestamps for whenever the chain is evoked, meaning whenever a question is fed to the chatbot.

julius-heitkoetter:
Fixed the issues requested. Here is an example:

[conversations_test.json](https://github.com/mit-submit/A2rchi/files/13032688/conversations_test.json)

ludomori99 (approved):
Looks great

PR #109, opened 2023-10-18
adding archi code to sources in meta archi

ludomori99:
added archi code as source to be used by meta archi


julius-heitkoetter (commented):
You may also want to change the prompts to reflect two things:

1. A2rchi is talking about itself
2. The context is code from the a2rchi repo

mdr223 (approved):
LGTM

julius-heitkoetter:
We should wait until #119 is resolved to do this, since right now I believe meta A2rchi doesn't even have a data_manager associated with it, meaning that it can't even really parse documents.

mdr223:
FYI I can work to get #119 merged after my DB class ends at 4pm

PR #111, opened 2023-10-18
Feature/code rendering

ludomori99:
Does not work as intended right now, so please not merge to main. Likely I am not understanding something about the css configuration.


julius-heitkoetter (approved):
LGTM, thanks for making the changes!

PR #113, opened 2023-10-19
Quick/move dev to submit06

ludomori99:
move dev deployment to submit06


mdr223:
Fwiw I believe you may be able to copy a docker volume from one machine to
another if the need ever arises.

PR #119, opened 2023-10-22
Feature/simplify services

mdr223:
I made two primary changes in this branch:
1. I moved the vectorstore data management functionality into the ChatWrapper
2. I renamed the data manager service to be the uploader service

By moving the vectorstore data management into the ChatWrapper we eliminate the need to have a long-running thread in the (formerly) data manager service which would fail silently. Now, that service only runs the uploader app, so I renamed it the uploader service.


julius-heitkoetter (changes requested):
I think we want to change the cleo interface wrapper class to also initialize the datamanager [(here)](https://github.com/mit-ubmit/A2rchi/blob/e912e89069d42948d538b04600d07daf82d49a0c/A2rchi/interfaces/cleo.py#L20) and then update it every call, just like the chat wrapper. This is so that we can maintain consistency of where the data manager is updating.

Also, as a bonus, we may want to create an parent wrapper class which has call and init methods which already do this. This is more of a formality and may make things sloppier than they need to be, but just a suggestion since these datamanager calls will be the same across all __init__ and __call__ methods in the wrapper class.

mdr223:
@julius-heitkoetter I think I've made the minimal necessary changes to the Cleo service to similarly enable us to perform data management outside of the uploader service.

I agree that we will likely want to further consolidate our AI wrapper classes (for Cleo and Chat) into a single (abstract) parent wrapper class, but I've decided to defer that to future work so that we don't hold up a chain of PRs.

julius-heitkoetter (approved):
Looks good!

PR #120, opened 2023-10-25
up query limit and fail gracefully if history is empty

mdr223:
See title.


julius-heitkoetter (approved):
Looks good! Could you make sure that there is an issue to remove this hotfix? Thanks!

PR #124, opened 2023-10-26
prepping release for submit06

mdr223:
see title


mdr223:
FYI, I have also copied over the `~/801-content` directory onto `submit06:~/801-content/`, so the volume mount for the data manager service should pick those files up.

julius-heitkoetter (approved):
LGTM

PR #129, opened 2023-10-26
move lock off of critical path containing call to OpenAI

mdr223:
This PR removes our lock from the call to OpenAI in the ChatWrapper, which should allow for concurrent requests to OpenAI.

Note that I did not narrow the scope of the lock on the like/dislike methods. This is because these methods do not take up much time, and I think it would be better to have clear lock acquisitions/releases on these methods rather than make changes with the risk of introducing a bug.

### DIFF
--- a/A2rchi/interfaces/chat_app/app.py
+++ b/A2rchi/interfaces/chat_app/app.py
@@ -15,7 +15,8 @@ import yaml
 import time
 
 # DEFINITIONS
-QUERY_LIMIT = 1000 # max number of queries 
+# TODO: remove this logic and eventually replace with per-user (or per-conversation) rate limits
+QUERY_LIMIT = 10000 # max number of queries 
 
 
 class ChatWrapper:
@@ -73,7 +74,9 @@ class ChatWrapper:
             print(" INFO - json_file found.")
 
         except FileNotFoundError:
+            # create data path if it doesn't exist
             print(" ERROR - json_file not found. Creating a new one")
+            os.makedirs(data_path, exist_ok=True)
 
         # update or add discussion
         discussion_dict = data.get(str(discussion_id), {})
@@ -93,9 +96,6 @@ class ChatWrapper:
         
         data[str(discussion_id)] = discussion_dict
 
-        # create data path if it doesn't exist
-        os.makedirs(data_path, exist_ok=True)
-
         # write the updated JSON data back to the file
         with open(os.path.join(data_path, json_file), 'w') as f:
             json.dump(data, f)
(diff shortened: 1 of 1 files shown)


julius-heitkoetter:
I there is another lock [here](https://github.com/mit-submit/A2rchi/blob/6239998b74a3683e8f439a3a4703a86e156653b0/A2rchi/chains/chain.py#L138). I’m pretty sure this is not used anymore now that we’re doing everything in the interface. However, we need to think carefully if the locking should take place at the interface level or the chain level. (i.e. I think locks that protect things happening in the chain should be kept in the chain). However, I think that our chain actually should be able to have multiple calls in parallel. So maybe we can just delete all the locks in the chain.

A2rchi/interfaces/chat_app/app.py
```diff
@@ -105,63 +105,79 @@ def __call__(self, history: Optional[List[Tuple[str, str]]], discussion_id: Opti
...
-            if score < similarity_score_reference and source in sources.keys(): 
-                output = "<p>" + result["answer"] + "</p>" + "\n\n<br /><br /><p><a href= " + sources[source] + ">Click here to read more</a></p>"
-            else:
-                output = "<p>" + result["answer"] + "</p>"
+            self.lock.release()
+            print("INFO - released lock file update vectorstore")
+        except Exception as e:
+            print(f"ERROR - {str(e)}")
+        finally:
+            self.lock.release()
```
ludomori99, line 120: Are we sure about the double release? I remember we discussed this already, but from what I saw it should be avoided to release locks already released. e.g. see in:   https://www.bogotobogo.com/python/Multithread/python_multithreading_Synchronization_Lock_Objects_Acquire_Release.php (can Ctrl+F "The release() method should only be called in the locked state")

ludomori99: ibid in update_discussion()

mdr223: Just updated a proposed fix

A2rchi/interfaces/chat_app/app.py
```diff
@@ -105,63 +105,77 @@ def __call__(self, history: Optional[List[Tuple[str, str]]], discussion_id: Opti
         """
         Execute the chat functionality.
         """
-        self.lock.acquire()
-        print("INFO - acquired lock file")
         try:
             # update vector store through data manager; will only do something if new files have been added
+            self.lock.acquire()
```
julius-heitkoetter, line 110: This self.lock.acquire should get moved out of the try, in front of it. This way, if something weird happens and we fail before/during execution of line 110, we won't silently fail.

mdr223: Good catch, just fixed

julius-heitkoetter (approved):

PR #130, opened 2023-10-27
Feature/postgres db

mdr223:
An initial PR for swapping `conversations_test.json` --> a Postgres database featuring two tables.

This PR is not completely finished, as I am still debugging / checking that one corner case is not breaking any assumptions made in the table construction, but I am putting it up now so that folks can get familiar with the contours of the code change. (Also note that I've only gone and changed the `dev` deployment, this way we can test things out in `dev` before also making these changes in e.g. `prod`, `prod-801`, etc., which will also require a migration from `conversations_test.json` into the new DB's that are put in place).

At a high-level, I've created two tables:
1. `conversations`
2. `feedback`

The schemas for these tables can be found in `deploy/dev/dev-init.sql`. The main change(s) occur in the `ChatWrapper` class, but there is some work that needs to be done to resolve this branch w/recent changes that were made to `main` following our Wednesday night fixes.


ludomori99 (commented):
Looks good. A couple things that came to my mind.
 1. Are we sure about putting the database functionalities (insert_feedback etc.) all inside the ChatWrapper? It seems to me that given we might want to extend the scope of the database (say to store logs or calls deeper in the chain, like inputs-outputs of the llm) we might want to have a separate class to manage the interactions with postgres. Just as a suggestion.
 2. Some `print` calls in app.py might lack context once we will look at the logs. Maybe some where for testing (249-256)? Or we could add text output to give context if we want.

.github/workflows/dev-ci-cd.yaml
```diff
@@ -87,31 +90,25 @@ jobs:
       # stop any existing docker compose that's running
       - name: Stop Docker Compose
         run: |
-          ssh submit-t3desk 'bash -s' < ${{ github.workspace }}/deploy/dev/dev-stop.sh
+          ssh submit06 'bash -s' < ${{ github.workspace }}/deploy/dev/dev-stop.sh
```
julius-heitkoetter, line 93: Why the move to subMIT06? I'm not against it, just want to make sure that we're consciously changing to other machines.

A2rchi/interfaces/chat_app/app.py
```diff
@@ -258,21 +392,24 @@ def like(self):
...
-            discussion_id = data.get('discussion_id')
             message_id = data.get('message_id')
 
             feedback = {
-                "chat_content" :  chat_content,
-                "message_id"   :  message_id,
-                "feedback"     :  "like",
+                "message_id"   : message_id,
+                "feedback"     : "like",
+                "feedback_ts"  : datetime.now(),
```
julius-heitkoetter, line 400: Why use datetime instead of time.time()? Just want to make sure we stay consistent across the repo with how we talk about timestamps.

A2rchi/interfaces/chat_app/app.py
```diff
@@ -177,14 +296,26 @@ def __call__(self, history: Optional[List[Tuple[str, str]]], discussion_id: Opti
             else:
                 output = "<p>" + self.format_code_in_text(result["answer"]) + "</p>"
 
-            ChatWrapper.update_or_add_discussion(self.data_path, "conversations_test.json", discussion_id, discussion_contents = history + [("A2rchi", output)])
+            # write user message and A2rchi response to database
+            user_message = (sender, content, msg_ts)
+            a2rchi_message = ("A2rchi", output, datetime.now())
+            print(user_message)
```
julius-heitkoetter, line 302: I agree with @ludomori99 that print statements such as these need at least minimal context in them (this just needs to be quick and dirty, we can swap to more robust logging later if needed).

julius-heitkoetter (commented):
Great work! I've added just a few small comments, but the only changed I'd request before merging would be the print statements. Everything else, I was just curious about.

julius-heitkoetter (approved):
LGTM

mdr223:
There was actually a somewhat involved merge conflict between this and the changes we made to move the lock off the critical path, I believe I've handled it correctly, but will request a second look.

julius-heitkoetter (commented):
LGTM

PR #132, opened 2023-10-27
Fix/dumb llm

ludomori99:
Was actually working already. The problem that was encountered earlier was due to the HuggingFaceEmbeddings apparently not working properly within the chromadb collections. PR containing only minor changes.


julius-heitkoetter (approved):
LGTM

Issue #134, opened 2023-10-30
HuggingFace embeddings unexpected behavior


ludomori99:
Two different error logs showed up (I was unable to understand the difference in the circumstance they'd arise).

Error 1: 
`name='dev_collection_with_HuggingFaceEmbeddings' id=UUID('a7b56b9e-7cf9-4cbc-b529-4c09c11fdc6f') metadata=None
Vectorstore needs to be updated
Files to remove: []
Files to add: {'391446804850.html': '/root/data/websites/391446804850.html', '311908320295.html': '/root/data/websites/311908320295.html', '299931406278.html': '/root/data/websites/299931406278.html', '948537108847.html': '/root/data/websites/948537108847.html', '213679292299.html': '/root/data/websites/213679292299.html', '198006956929.html': '/root/data/websites/198006956929.html', '247039305341.html': '/root/data/websites/247039305341.html', '136901283757.html': '/root/data/websites/136901283757.html', '649811655156.html': '/root/data/websites/649811655156.html', '178898586411.html': '/root/data/websites/178898586411.html', '305701533622.html': '/root/data/websites/305701533622.html', '243114892764.html': '/root/data/websites/243114892764.html', '181001073821.html': '/root/data/websites/181001073821.html', '147197139266.html': '/root/data/websites/147197139266.html', '149446187468.html': '/root/data/websites/149446187468.html', '191440252181.html': '/root/data/websites/191440252181.html', '100971305273.html': '/root/data/websites/100971305273.html'}
Ids:  ['391446804850159290159290', '391446804850159440159440', '391446804850493563493563', '391446804850333332333332', '391446804850874070874070']
Created a chunk of size 1383, which is longer than the specified 1000
Created a chunk of size 1229, which is longer than the specified 1000
Created a chunk of size 1266, which is longer than the specified 1000
Ids:  ['311908320295287408287408', '311908320295166536166536', '311908320295300922300922', '311908320295462437462437', '311908320295235417235417', '311908320295272625272625', '311908320295284394284394', '311908320295223642223642', '311908320295150568150568', '311908320295337206337206', '311908320295185710185710', '311908320295149215149215', '311908320295374140374140', '311908320295352770352770', '311908320295384027384027', '311908320295161701161701', '311908320295186736186736', '311908320295164042164042', '311908320295270457270457', '311908320295804874804874', '311908320295343918343918', '311908320295154588154588', '311908320295289953289953']
Traceback (most recent call last):
[...]

PR #140, opened 2023-11-03
Changed HuggingFace embedding model

ludomori99:
Various small changes.
- changed the model from the sentence-transformers library. Appears to have slightly worse performance but is much faster. Reason for change is it is the only one among the ones tested (and state-of-the-art) that does not produce the bug "collection not found" #134 . Root deployment now uses open source embedding and includes also full tutorial code in the ragging docs. 
- activate/deactivate copy button in code panels from config. Keep deactivated as long as we don't have https connection
- Leave ALLOW_RESET environment variable in deploy script. Is not directly needed at the moment, but given the time spent figuring out how to allow the vectorstore client to reset I figured it might save time in future debugging to keep it there in the dev. The reset command has been used in debugging but now is removed.


mdr223 (approved):
Approving as this all looks good to me, just wanted to check and see if you meant to use `ALLOW_RESET` anywhere? I saw it was added to the compose file but I didn't see it being used in the code.

mdr223:
Ahhh I see, makes sense and thanks for the explanation!

PR #141, opened 2023-11-07
Feature/add timeout to client requests

mdr223:
This feature adds timeout functionality to the client and server-sides of our chat application. On the client side we timeout the request after a specified duration (default 120s). Once the timeout occurs, the client will display the "Oops!" error message.

We also need to handle request timeouts at the server. Simply timing out a request from the client does not stop the server from trying to process the request (and calling OpenAI), only to produce a result that will be in vain. Thus, right before we call the Chain, we check to see if the timeout has already occurred. If so, we return an error instead of executing the request.

I also made some minor changes to the front-end:
- Updated the CSS to add the same padding for "Oops!" messages as we have in regular messages (previously, the "Oops!" messages had no space between the A2rchi profile pic and the start of the text, which looked bad)
- I added a `try-catch` block around our `fetch` calls for `likeResponse` and `dislikeResponse`, as these methods will currently fail if/when a user likes/dislikes an "Oops!" message.

### DIFF
--- a/.github/workflows/dev-ci-cd.yaml
+++ b/.github/workflows/dev-ci-cd.yaml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - feature/add-timeout-to-client-requests
 jobs:
   deploy-dev-system:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


julius-heitkoetter (approved):
LGTM

PR #181, opened 2024-12-02
Cleanup/october merge

julius-heitkoetter:
This is merging into main all the changes from the october cleanup. 

The main change is an adjustment to how we do deployments, from a CI/CD github workflow to a easy to install and use cli.

For more information, see all PRs associated with the `cleanup/october_merge` branch as well as PR #159, which was used as a base for this branch.


julius-heitkoetter:
oops, this is not the way to do this, since we don't want a merge. We want a full FF/rebase.

PR #195, opened 2025-04-01
hf model support

pmlugato:
- support for any huggingface model
- option --gpu to give container access to and run on gpu
- option --podman if using podman instead of docker
- fixed handling of chat histories which weren't getting passed correctly in open-source models
- some more logging
- model caching to avoid loading into memory twice
- some extra stuff on the chatbot website (in particular some waiting messages for the longer response times with open source models)
- minor bug fixes

### DIFF
--- a/a2rchi/templates/dockerfiles/Dockerfile-cleo-gpu
+++ b/a2rchi/templates/dockerfiles/Dockerfile-cleo-gpu
@@ -1,0 +1,21 @@
+# syntax=docker/dockerfile:1
+FROM docker.io/pytorch/pytorch:2.2.0-cuda12.1-cudnn8-devel
+RUN mkdir -p /root/A2rchi
+WORKDIR /root/A2rchi
+
+COPY requirements.txt requirements.txt
+COPY LICENSE LICENSE
+
+RUN pip install --upgrade pip && pip install -r requirements.txt
+
+RUN apt-get update && apt-get install -y \
+    nvidia-utils-550 \
+    libnvidia-compute-550 \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY a2rchi_code a2rchi
+COPY pyproject.toml pyproject.toml
+COPY weblists weblists
+RUN pip install --upgrade pip && pip install .
+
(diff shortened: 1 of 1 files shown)


a2rchi/chains/chain.py
```diff
@@ -44,8 +44,16 @@ def update_config(self):
         condense_model_name = self.chain_config.get("CONDENSE_MODEL_NAME", model_name)
         summary_model_name = self.chain_config.get("SUMMARY_MODEL_NAME", model_name)
         self.llm = model_class_map[model_name]["class"](**model_class_map[model_name]["kwargs"])
-        self.condense_llm = model_class_map[condense_model_name]["class"](**model_class_map[condense_model_name]["kwargs"])
-        self.summary_llm = model_class_map[summary_model_name]["class"](**model_class_map[summary_model_name]["kwargs"])
+
+        if condense_model_name == model_name:
```
mdr223, line 48: On lines 44-45, is it possible for `self.chain_config.get("CONDENSE_MODEL_NAME", model_name)` to return the same model as `model_name`? I.e. could the user set a `CONDENSE_MODEL_NAME` that is the same as `model_name`, which would then cause this statement to set `self.condense_llm` to `None` even though it should be set to `model_name`?

If this isn't an issue, feel free to ignore this comment. Otherwise, maybe do `condense_model_name = self.chain_config.get("CONDENSE_MODEL_NAME", None)` and then in this if statement you can check `if condense_model_name is not None`.

(same logic applies for `summary_model_name`.

mdr223 (approved):
Left one minor comment, otherwise LGTM!

a2rchi/chains/chain.py
pmlugato: tldr; this helps to remove an issue where the same model was getting loaded into memory twice: one for base llm, one for condense.

The logic here is that `self.condense_llm = None` (summary too) unless the user specifies in the config that they want use a different model for condensing. If None, in `BaseChain.from_llm(…)` (see base.py) from the method `update_vectorstore_and_create_chain` below, it will just take the base LLM and use it for the condensing chain as well.

PR #200, opened 2025-06-04
Add Base SSO and CERN specific SSO Scraper

hassan11196:
Adds support to crawl pages behind CERN SSO.  and provides Base class that can be used to extend for other SSO systems.

### Update your base_config as follows:
```
utils:
  sso:
    ENABLED: True
    SSO_CLASS: CERNSSOScraper
    SSO_CLASS_MAP:
      CERNSSOScraper:
        kwargs:
          headless: True # headless mode for scraping, as it is run in a container
          max_depth: 2 # max depth for crawling from starting URL

```
### Add the URLs behind sso by appending them with `sso-` in the list files:
```
sso-https://cmspnr.docs.cern.ch/
```

### you would need a cern service account for authentication:
1. Place your account username in `sso_username.txt` and your password in `sso_password.txt` in the secrets folder.
Note: CERN personal accounts are now 2FA secure and they will not work. you can create a cern service account for this purpose.

### DIFF
--- a/a2rchi/cli/cli_main.py
+++ b/a2rchi/cli/cli_main.py
@@ -519,8 +519,11 @@ def create(
 
     _prepare_secret(a2rchi_name_dir, "pg_password", locations_of_secrets)
     # SSO secrets
-    _prepare_secret(a2rchi_name_dir, "sso_username", locations_of_secrets)
-    _prepare_secret(a2rchi_name_dir, "sso_password", locations_of_secrets)
+    if a2rchi_config.get("utils",{}).get("sso", {}).get("ENABLED", False):
+        _print_msg("Preparing SSO secrets")
+        compose_template_vars["sso"] = True
+        _prepare_secret(a2rchi_name_dir, "sso_username", locations_of_secrets)
+        _prepare_secret(a2rchi_name_dir, "sso_password", locations_of_secrets)
 
 
     # copy prompts (make this cleaner prob)
--- a/a2rchi/templates/base-compose.yaml
+++ b/a2rchi/templates/base-compose.yaml
@@ -12,8 +12,10 @@ services:
       postgres:
         condition: service_healthy
     environment:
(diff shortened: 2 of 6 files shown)


a2rchi/cli/cli_main.py
```diff
@@ -439,6 +460,9 @@ def create(
     shutil.copyfile(a2rchi_config["chains"]["prompts"]["CONDENSING_PROMPT"], os.path.join(a2rchi_name_dir, "condense.prompt"))
     shutil.copyfile(a2rchi_config["chains"]["prompts"]["SUMMARY_PROMPT"], os.path.join(a2rchi_name_dir, "summary.prompt"))
 
+    # TODO: remove this in favor of config file and get sso creds from secrets
+    print("Copying .env file to a2rchi directory", a2rchi_name_dir)
+    shutil.copyfile('./.env', os.path.join(a2rchi_name_dir, ".env"))
```
pmlugato, line 465: I think we've discussed we should maybe look at how we handle secrets to make things a bit cleaner in general, but  for now can you have some documentation somewhere of what is needed in this file for the scraper to work? 

just a small section to explain and then in time when we update the guide and also secrets we can update as needed.

hassan11196: Hi @pmlugato, 
I have updated the scraper to use the secrets similar to how we handle other secrets for now, As we already discussed that the secrets management needs to be revamped, I will create a seperate PR for it.

hassan11196:
How to set up SSO credentials, 
In your secret folder, create two files `sso_username` and `sso_password` with your username and password for the SSO account, respectively. 2FA-secured accounts will not work.

a2rchi/cli/cli_main.py
pmlugato: Requested last change to ensure cli doesn't fail out when sso secrets aren't provided, lest the service is requested.

PR #212, opened 2025-06-24
Piazza service

pmlugato:
piazza service rewritten and reorganized a bit

### DIFF
--- a/a2rchi/interfaces/piazza.py
+++ b/a2rchi/interfaces/piazza.py
@@ -1,0 +1,151 @@
+from a2rchi.chains.chain import Chain
+from a2rchi.interfaces.uploader_app.app import FlaskAppWrapper
+from a2rchi.utils.config_loader import Config_Loader
+from a2rchi.utils.data_manager import DataManager
+from a2rchi.utils.env import read_secret
+from a2rchi.utils.scraper import Scraper
+
+from flask import Flask
+from piazza_api import Piazza as PiazzaAPI
+from threading import Thread
+
+import json
+import os
+import requests
+import time
+
+
+class PiazzaAIWrapper:
+    def __init__(self):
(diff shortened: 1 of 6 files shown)


a2rchi/interfaces/piazza.py
```diff
@@ -0,0 +1,152 @@
...
+        self.piazza.user_login(email=self.piazza_email, password=self.piazza_password)
+        self.piazza_net = self.piazza.network(self.piazza_config["network_id"])
+
+        # slack webhook for sending draft responses
+        self.slack_url = read_secret("SLACK_WEBHOOK")
+        self.slack_headers = {'content-type': 'application/json'}
+
+        # 
+        self.min_next_post_file = "/root/data/min_next_post.json"
+        self.default_min_next_post_nr = 500 # should default to most recent post in the piazza + 1
```
mdr223, line 64: If desired, you could copy some of the `dynamic_min_next_post_nr` logic from below to avoid setting this arbitrary constant; but it seems your logic below will gracefully handle the case where e.g. 500 is greater than the next post number, so this is just a suggestion.

mdr223 (approved):
Left one suggestion, but otherwise it LGTM

a2rchi/interfaces/piazza.py
pmlugato: Yeah, read_min_next_post() should be pretty robust as far as getting the most recent post, but I put this as just an extra fallback. However, maybe it would be better instead of choosing arbitrary constant to just remove this and error out if there are no posts/it can't find the most recent one? Because as you say 500 was arbitrary, and if it does get set and there are only 10 posts, simply nothing will happen until reaching 500 posts (or if I set default to 1 and there are 500 posts, then the slack will get spammed...)

PR #213, opened 2025-06-24
Integrate JIRA into a2rchi

haozturk:
### Usage
1. Define a JIRA_URL and JIRA_PROJECTS in the config, e.g.
```
# Rest of the config
interfaces:
  jira:
    JIRA_URL: "https://its.cern.ch/jira"
    JIRA_PROJECTS: ['CMSPROD']
    ANONYMIZE_DATA: True
# Rest of the config
```
2. Put the JIRA secret (Personal Access Token) in the usual secret folder. Its name should be `jira_pat.txt`
3. Create the deployment with `--jira True`
```
$ a2rchi create --name compops-a2rchi --a2rchi-config configs/comp_ops/base_config.yaml --jira True
```
This will make a2rchi fetch all the JIRA data, put it in the data folder and index them. 

image

it works as it can be seen in the screenshot

### Few concerns, future todos:

1. Since initially I wasn't planning to make this interactive, I didn't create a new service and image for this. So, it currently lives in the chat app container. We can move it to its own service either in this PR or later on if that's preferable.
2. I haven't deeply investigated anonymization. I checked a sample of 10 tickets and it worked as expected for those.
3. Currently, we're getting issue id, creation date, title, body and comment. Other information such as date of comments and priority can be useful to add later.
4. We should make a2rchi capable of responding to JIRA tickets in the future.

@lucalavezzo @pmlugato @hassan11196

### DIFF
--- a/a2rchi/utils/anonymizer.py
+++ b/a2rchi/utils/anonymizer.py
@@ -1,0 +1,67 @@
+"""
+Authors: Pietro Lugato, Hasan Ozturk
+"""
+
+import re
+import spacy
+from typing import Set, List
+
+
+class Anonymizer:
+    def __init__(self, excluded_words: Set[str] = None):
+        """
+        Initialize the Anonymizer.
+        """
+        try:
+            self.nlp = spacy.load("en_core_web_sm")
+        except OSError:
+            spacy.cli.download("en_core_web_sm")
+            self.nlp = spacy.load("en_core_web_sm")
+        # TODO: Move the following patters into the config file
+        self.EXCLUDED_WORDS = excluded_words or {"John", "Jane", "Doe"}
+        self.GREETING_PATTERNS = [
+            re.compile(r"^(hi|hello|hey|greetings|dear)\b", re.IGNORECASE),
+            re.compile(r"^(\w+,\s*|\w+\s+)", re.IGNORECASE),
+        ]
+        self.SIGNOFF_PATTERNS = [
+            re.compile(r"\b(regards|sincerely|best regards|cheers|thank you)\b", re.IGNORECASE),
+            re.compile(r"^\s*[-~]+\s*$"),
+        ]
+        self.email_pattern = re.compile(r"[\w\.-]+@[\w\.-]+\.\w+")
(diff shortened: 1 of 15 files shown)


pmlugato (approved):

a2rchi/utils/anonymizer.py
```diff
@@ -0,0 +1,68 @@
...
+from typing import Set, List
+
+
+class Anonymizer:
+    def __init__(self, excluded_words: Set[str] = None):
+        """
+        Initialize the Anonymizer.
+        """
+        try:
+            self.nlp = spacy.load("en_core_web_sm")
```
hassan11196, line 16: if possible can this NER model can also be moved to the a2rchi configurable similar to how we do it for the embedding models. i.e 
https://github.com/mit-submit/A2rchi/blob/fd3c74794883f42874c87549eb7db4df18fddbaa/a2rchi/templates/base-config.yaml#L125

a2rchi/utils/anonymizer.py
```diff
@@ -0,0 +1,68 @@
...
+        except OSError:
+            spacy.cli.download("en_core_web_sm")
+            self.nlp = spacy.load("en_core_web_sm")
+        # TODO: Move the following patters into the config file
+        self.EXCLUDED_WORDS = excluded_words or {"John", "Jane", "Doe"}
+        self.GREETING_PATTERNS = [
```
hassan11196, line 22: I see the todo, so if you move the NER model to the config file, this can be placed in the same hierarchy as well

hassan11196 (changes requested):
Looks good, a small config change in the Anonymizer to ease change in NER models

a2rchi/utils/anonymizer.py
hassan11196: Looks good Thanks

a2rchi/utils/anonymizer.py
hassan11196: LGTM

hassan11196 (approved):
Thank you for the changes, LGTM.

PR #224, opened 2025-07-08
[WIP] Support VLLM

lucalavezzo:
- Add support for VLLM models
- Update requirements.txt to include updated packages, and `vllm`
- input files are not required to launch the chatbot
- add `mkdocs` to packages to compile docs, and update a2rchi package authors

### DIFF
--- a/a2rchi/cli/cli_main.py
+++ b/a2rchi/cli/cli_main.py
@@ -281,7 +281,7 @@ def create(
     required_fields = [
         'name', 
         'global.TRAINED_ON',
-        'chains.input_lists', 
+        #'chains.input_lists', 
         'chains.prompts.CONDENSING_PROMPT', 'chains.prompts.MAIN_PROMPT', 'chains.prompts.SUMMARY_PROMPT',
         'chains.chain.MODEL_NAME', 'chains.chain.CONDENSE_MODEL_NAME', 'chains.chain.SUMMARY_MODEL_NAME'
     ]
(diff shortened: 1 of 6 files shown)


a2rchi/cli/cli_main.py
```diff
@@ -281,7 +281,7 @@ def create(
     required_fields = [
         'name', 
         'global.TRAINED_ON',
-        'chains.input_lists', 
+        #'chains.input_lists',
```
pmlugato, line 284: If this doesn't fail elsewhere, you can just remove the line, don't need to leave commented out. Then maybe add a warning that if `chains.input_lists` is empty, no documents are being passed *at deployment* (since maybe some docs are still in corresponding volume) for the RAG system to pull from. Also, this should be reflected in user's guide.

lucalavezzo: Removed comment. Indeed I am sure the docs could be improved...

requirements.txt
```diff
@@ -44,16 +44,16 @@ langchain-community==0.3.0
...
 langchain-huggingface==0.1.0
-pydantic==2.9.2
-pydantic-core==2.23.4
-pydantic-settings==2.5.2
+pydantic==2.11.7
+pydantic-core==2.33.2
 chroma-hnswlib==0.7.3
 anthropic==0.34.2
-torch==2.2.0
+torch==2.7.0
```
pmlugato, line 51: Other than vLLM, did you also test the HuggingFaceOpenLLM class -- which uses the transformers package -- to make sure things still work with this version of torch?

PR #225, opened 2025-07-09
introduce new parameter to switch verification of URLs in the uploader

Markus-Paus:
should not interfere with anyone else, very low risk.


pmlugato (approved):
looks good to me!

PR #226, opened 2025-07-09
Fixing the boolean flag updates.

Markus-Paus:
Oopsie


Markus-Paus:
There was an additional bug I had to fix for the grafana user creation in postgres

PR #239, opened 2025-07-30
VLLM update, version 1

z4027163:
The VLLM update:
1. Adding important variables to base-config
2. Adjusting VLLM version to older one 0.8.5, since the newer version has bug leading to: TypeError: XFormersImpl.__init__() got an unexpected keyword argument 'layer_idx' .
      Related details: https://github.com/vllm-project/vllm/issues/20634 
3. A new requirements.txt file is created: requirements_NLLV8.txt,  make sure to use this one when using VLLM.
4. Add prompt reformat to remove HTML related strings
5. Add truncate function to use up to 10 user/A2rchi exchanges within the length limit (max_model_len).
6. Change base model default to Qwen. So far the VLLM doesn't work for deepseek due to its thinking process. 
7. Enable multi-GPU for VLLM. This requires to change mp to "spawn", which need to change the structure of service_chat.py


a2rchi/chains/models.py
```diff
@@ -204,49 +207,123 @@ def _call(
...
+            )
             self.set_cached_model(model_cache_key, (None, self.vllm_engine))
 
+        print(f"[VLLM] input nGPU={self.tensor_parallel_size}")
+
     @property
     def _llm_type(self) -> str:
         return "custom"
 
+    def _truncate_prompt(self, prompt: str) -> str:
```
pmlugato, line 289: for limiting the history to the last 10 interactions, this shouldn't be done here, but I agree this is a useful thing to implement while better chat history handling is being implemented -- could you move this logic to the place where this formatted history is being constructed, the `_get_chat_history` function in `a2rchi/chains/base.py`. 

Later, it can be added to the TokenLimiter class, but Luca is working on the history handling in general so will probably change/move this around anyway so doesn't need to be in this PR

a2rchi/templates/dockerfiles/Dockerfile-chat
```diff
@@ -33,4 +33,7 @@ COPY pyproject.toml pyproject.toml
 COPY weblists weblists
 RUN pip install --upgrade pip && pip install .
 
-CMD ["python", "-u", "a2rchi/bin/service_chat.py"]
\ No newline at end of file
+RUN sed -i 's/AutoConfig.register("aimv2", AIMv2Config)/AutoConfig.register("aimv2", AIMv2Config, exist_ok=True)/' \
+    /opt/conda/lib/python3.10/site-packages/vllm/transformers_utils/configs/ovis.py
+
```
pmlugato, line 38: Should this only be in the `-gpu` Dockerfile?

requirements_VLLM8.txt
pmlugato, line 1: I think it is not a good idea to have two requirements files, even if just temporarily. Can you update the `requirements.txt` with the newer version of vLLM, which I think you had done, and then any specific models that you saw weren't working with this newer version because of a bug that was introduced (like I think you saw with some `Qwen/` models, then we just don't support those models until is it fixed.

Let me know if this is reasonable or if I misunderstood and there is really a fundamental mismatch between some packages...

z4027163: For newer version of VLLM: 
1. The multi-gpu doesn't work, due to NCCL related issue. 
`RuntimeError: NCCL error: unhandled system error (run with NCCL_DEBUG=INFO for details)`
I have no estimation how long it will take to resolve this, but at least it will more than 1 week. 

2. All the Qwen models probably doesn't work. All the thinking models doesn't work, as its thinking procedure would be treated as rounds of user asking questions. So far I have not found a successful model that could work under the newer VLLM version. 

So far upload the requirements_v10.txt as for now, including v0.10.0 vllm (latest version) with updated dependent packadges

a2rchi/chains/models.py
```diff
@@ -256,19 +333,32 @@ def _call(
...
         )
 
         # vLLM expects a list of prompts
-        outputs = self.vllm_engine.generate([prompt], sampling_params)
+        outputs = self.vllm_engine.generate([formatted_prompt], sampling_params)
+        #outputs = self.vllm_engine.generate([prompt], sampling_params)
         # outputs is a list of RequestOutput, each has .outputs (list of generations)
         if outputs and outputs[0].outputs:
+            #text = outputs[0].outputs[0].text
+            #return text.replace("</think>", "").strip()
```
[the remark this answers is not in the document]
z4027163: extra line, this is cleaned up now.

requirements_VLLM8.txt
z4027163: After some checks, the multi-GPU can not be resolved in a short time. So far no firm solution is found online.

PR #240, opened 2025-07-30
Improving & uniformizing logging; reformat QA chain for chatbot

pmlugato:
*logging*:
- `a2rchi.utils.logging` defines basicConfig for python's logging package
- `--debug` option available in CLI to set level to logging.DEBUG (default is INFO)
- function `setup_logging` from `a2rchi.utils.logging` should be called at beginning of executable before apps run (e.g. in `a2rchi/bin/service_chat.py`)
- at the beginning of different scripts and files, `from a2rchi.utils.logging import get_logger` and set the logger, `logger = get_logger(__name__)` -- debug mode adds the module name to the format to more easily see where things are happening
- also find in the container at `/root/data/chain_input_output.log` or in corresponding volume (e.g., `/var/tmp/<user>/containers/storage/volumes/a2rchi-<deployment name>/_data/chain_input_output.log` (filename can be changed in config at `chains:base:logging:input_output_filename`)) the fully templated prompts to both the question-condensing LLM and the main (QA) LLM, with timestamps

*BaseSubMITChain --> BaseQAChain*:
- rewrote the base class for the QA chain (no longer used only for SubMIT)
- moved away from deprecated langchain classes and methods
- tried to expose a bit more of what goes on in the chain as opposed to leaving it to hooks in parent langchain class (as was done previously)
- also made the TokenLimiter (`a2rchi.chains.utils.token_limiter`) a bit more robust when it is unable to check the LLM's max tokens attribute (if it doesn't exist or is a bad value).


a2rchi/utils/logging.py
```diff
@@ -0,0 +1,22 @@
+import logging
+
+# ignore debug logs from these modules, too verbose :)
+ignore_debug_modules = ["urllib3.connectionpool", "filelock"]
+
+def setup_logging(debug=False):
+
+    if debug:
```
lucalavezzo, line 8: could in principle support -v=0,1,2,3,4 for the different levels of logging verbosity instead of only INFO and DEBUG, but I leave it to you whether it's worth the change now

a2rchi/chains/utils/token_limiter.py
```diff
@@ -17,9 +20,18 @@ def reduce_tokens_below_limit(self, docs: list[Document]) -> list[Document]:
...
 
-        tokens = [
-            self.llm.get_num_tokens(doc.page_content) for doc in docs
-        ]
+        tokens = []
+        for i, doc in enumerate(docs):
+            try:
+                doc_tokens = self.llm.get_num_tokens(doc.page_content)
+                if doc_tokens is None or doc_tokens < 0:
+                    doc_tokens = max(len(doc.page_content) // 4, 1)
```
lucalavezzo, line 28: would be cleaner to error out here so as to go to the Exception, which would reduce code duplication

lucalavezzo (approved):
Thanks a lot for the PR. I will approve, my suggestions are only for code cleanliness, not functionality. Your changes to chains/base.py I think will be really useful, the code is already much more readable.

a2rchi/bin/service_chat.py
```diff
@@ -2,19 +2,23 @@
...
 from a2rchi.utils.env import read_secret
+from a2rchi.utils.logging import setup_logging
 
 from flask import Flask
 
 import os
 
+# set basicConfig for logging
+debug = Config_Loader().config["debug"]
+setup_logging(debug)
```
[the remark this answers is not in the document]
pmlugato: Moved

a2rchi/utils/logging.py
pmlugato: Changed to this, indeed I think makes more sense to just have this now.

pmlugato:
Can be a separate PR, but also noting that we should update the CLI as well to make those logs more readable.

a2rchi/bin/service_chat.py
```diff
@@ -41,4 +44,4 @@ def generate_script(config):
     template_folder=config["template_folder"],
     static_folder=config["static_folder"],
 ))
-app.run(debug=True, use_reloader=False, port=config["PORT"], host=config["HOST"])
+app.run(debug=debug, use_reloader=False, port=config["PORT"], host=config["HOST"])
```
haozturk, line 47: Running the app in debug mode is different than making the verbosity level DEBUG. In production, we might want to run the app debug=False while verbosity=DEBUG. At this stage of the project, I think it's not important as we run in debug mode by default anyway, but I can see that this can be a problem later on. Perhaps, make an issue to address it later?

haozturk (changes requested):
Thanks a lot for this, it was much needed. I think the format should be slightly updated if you agree as I pointed out inline. I will approve once we resolve that thread.

haozturk (approved):
Thanks for changes, looks good to me now. I'll approve it now. Feel free to merge it yourself once the conflicts are resolved.

PR #241, opened 2025-07-31
Mattermost branch

mariadalfonso:
Can activate the service with 
`a2rchi create --name name-a2rchi --a2rchi-config configs/name.yaml --mattermost True`

for now tested writing into town-square of our channel (to be fixed)
see 
https://github.com/mit-submit/A2rchi/compare/main...mariadalfonso:mattermostBranch?expand=1#diff-bf32b4c2221358f719ed62f77892c317c2711bc7219bd52a58e6da2321e7ca5dR77

### DIFF
--- a/a2rchi/cli/cli_main.py
+++ b/a2rchi/cli/cli_main.py
@@ -214,6 +214,7 @@ def cli():
 @click.option('--cleo-and-mailer', '-cm', 'include_cleo_and_mailer', type=bool, default=False, help="Boolean to add service for a2rchi interface with cleo and a mailer")
 @click.option('--jira', '-j', 'include_jira', type=bool, default=False, help="Boolean to add service for a2rchi interface with Jira")
 @click.option('--piazza', '-piazza', 'include_piazza_service', type=bool, default=False, help="Boolean to add piazza service to read piazza posts and suggest answers to a slack channel.")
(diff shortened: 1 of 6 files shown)


a2rchi/cli/cli_main.py
```diff
@@ -288,6 +290,9 @@ def create(
     if include_piazza_service:
         required_fields.append('utils.piazza.network_id')
 
+#    if include_mattermost_service:
```
pmlugato, line 293: Should this be commented or is it indeed required? If it is not needed or is old please remove

pmlugato:
I've added a few comments, then like you say once the logging is updated, it is good to merge as a start for this service. Two main things I've noted down for next update are (which I can make an issue with):

1. **GPU support**. Should be quick to implement, adding a new dockerfile with `-gpu` appended, similar to chat, changing the `base-compose.yaml`, and quickly restructuring `a2rchi/bin/service_mattermost.py` to reflect the structure in `a2rchi/bin/service_chat.py` so vLLM can run properly.
2. **Update handling of how a2rchi decides what to respond to**. It seems right now that a2rchi responds to any user message it has not already responded to, but in a chat there will also be other people that respond to questions or have some back and forth, i.e., messages that a2rchi probably shouldn't respond to. Not sure immediately the best way to handle this but something to think about...

Everything else looks good! Thanks for the work on this

PR #244, opened 2025-08-09
Chat history in BaseQA chain; revamped TokenLimiter; misc. framework improvements

lucalavezzo:
- add chat history to the QA chain
- expanded TokenLimiter to limit history and docs, prioritizing docs
- applied limiting of tokens to condensing prompt
- add protection for large user inputs, a2rchi fails gracefully, returning a warning
- updated submit.prompt to use chat history
- the max tokens passed to the LLM are now controlled via the config
- the size of the prompt is correctly subtracted from the max tokens to avoid underestimation
- Checking of input/output safety moved to its own utility 
- by default, `ConfigLoader` doesn't map the LLM names to the class to avoid circular imports and needless computation
- `ConfigLoader` --> `load_config()` no need for this to be a class


a2rchi/utils/config_loader.py
```diff
@@ -7,26 +7,26 @@
 # DEFINITIONS
 CONFIG_PATH = "/root/A2rchi/config.yaml"
 
-class Config_Loader:
+class ConfigLoader:
 
-    def __init__(self):
-        self.config = self.load_config()
+    def __init__(self, map: bool = True):
+        self.config = self.load_config(map=map)
```
pmlugato, line 13: While you are adding this, I think it makes sense to make map=False so that most loading of the config does not make unnecessary imports, and can be specified when needed (like in `a2rchi/utils/data_manager.py` and `a2rchi/chains/chain.py`, mainly).

PR #248, opened 2025-08-12
Git scraper

juanpablosalas:
Included GitScraper class in order to scrape git based documentation that uses MKDocs. This feature is inspired by CompOps documentation:

https://gitlab.cern.ch/cms-tier0-ops/documentation.git
https://gitlab.cern.ch/cmsdmops/Documentation.git
https://gitlab.cern.ch/CMSProductionReprocessing/documentation

### Update your base_config as follows:
```
utils:
  git:
    ENABLED: True
```
### Add the git HTTPS url by appending them with `git-` in the list files:
```
git-https://gitlab.cern.ch/cms-tier0-ops/documentation.git
```

Note that this must be MkDocks-based documentation repositories (i.e. contain the mkdocs.yml [file](https://www.mkdocs.org/getting-started/)).

### You would need a GitHub/GitLab username and token for authentication (read only should work fine):
Place your account username in `git_username.txt` and your token in `git_token.txt` in the secrets folder.


a2rchi/utils/git_scraper.py
```diff
@@ -0,0 +1,169 @@
...
+
+        # create sub-directory for git repositories if it doesn't exist
+        self.git_dir = os.path.join(self.data_path, "git")
+        os.makedirs(self.git_dir, exist_ok=True)
+
+        self.git_username = read_secret("GIT_USERNAME")
+        self.git_token = read_secret("GIT_TOKEN")
+
+        if self.git_username == '' or self.git_token=='':
+            FileNotFoundError("Git Personal Access Token (PAT) not found. Please set it up in your environment.")
```
pmlugato, line 29: What's the purpose of creating this exception object here? If I understand correctly, if "git" is enabled in the config, the CLI will enforce the secrets being there. If not, this class will still get initialized, and from my reading it seems nothing will break, but maybe you want to log something like "no git repos, skipping scraping" or something? Let me know if I am missing something.

juanpablosalas:
Thanks for checking @pmlugato!
I added the yaml path to the base config and included the configuration steps in the documentation as an optional configuration field (lmk if you think they should go elsewhere).

pmlugato (approved):
Thanks for the changes, Juan Pablo, I will merge now!

PR #251, opened 2025-08-12
SSO scraper fix and inclusion of URL reference.

Markus-Paus:
This is not the final version but will work for the CMS TWiki scrapes we are doing for now. Maybe we should change this all together, so maybe this is just good enough for now before a more complete overhaul happens.


hassan11196 (approved):
Thanks for the PR, I have tested the changes on our instance and they are good. Just a small comment on the verbose log statement

PR #262, opened 2025-08-19
Add health check

holzman:


holzman:
Yes, the plan is to have a health endpoint for a k8s deployment.
I’m agnostic as to the decorator vs manual approach.

- B


From: pmlugato ***@***.***>
Date: Tuesday, August 19, 2025 at 5:11 PM
To: mit-submit/A2rchi ***@***.***>
Cc: Burt Holzman ***@***.***>, Author ***@***.***>
Subject: Re: [mit-submit/A2rchi] Add health check (PR #262)

[EXTERNAL] – This message is from an external sender

@pmlugato approved this pull request.

I assume you are already using this in Kubernetes deployment or plan to add explicit healthcheck for chat service to the compose file? Also, I would follow the manual approach to adding endpoints just for consistency, unless there is reason to use decorator here instead?

I will approve and leave it to you to change if you agree, then feel free to merge it.

PR #263, opened 2025-08-19
Permissions for OpenShift/OKD

holzman:
OpenShift and OKD run pods as an ephemeral user with GID 0 rather than root.


pmlugato (approved):
Sounds good, merged.

PR #269, opened 2025-08-22
Introduce hybrid retrieval with BM25

haozturk:
This PR introduces BM25Retriever, a retriever capable of doing lexical search. It's accessible via a new class called `HybridRetriever` which calles both lexical and semantic retrieval and combines the results.

By default, hybrid search is disabled. To enable, one needs to configure it as follows:
```
data_manager:
  use_hybrid_search: True
```

Additionally, it's possible to configure the weight of the lexical and the semantic search as well as the settings of BM25/lexical search as follows:
```
data_manager:
  use_hybrid_search: True
  bm25_weight: 0.6
  semantic_weight: 0.4
  bm25:
    k1: 0.5
    b: 0.75
```

I want different deployments to try it out and then we can decide if we want to a2rchi use it by default.


haozturk:
Changes can be reviewed, but note that all the docs retrieved are removed by the token limiter due to a bug. I believe @lucalavezzo  is fixing that bug at https://github.com/mit-submit/A2rchi/pull/265

But since that PR is still in progress, I can include the fix for that bug in this PR if preferred. This way the feature can be fully tested (I can likely do it on Sunday or Monday)

pmlugato:
Hey @haozturk , from the quick look I have done, I definitely don't see anything major wrong and would be happy to approve!

With respect to the bug in the token limiter, I think the best course of action is to include the bug fix if it isn't too much, so that people can start playing with the hybrid retriever and testing like you say, and we can get some feedback. I am traveling and won't be able to fully review the PR from @lucalavezzo which has the fix, but also includes a lot of reorganization/restructuring (which by the way I encourage you to take a look at when you get the chance and also give some comments, I will request you there :).

I will be back later next week, but for now I will approve, and then once you include the token limiter bug fix, feel free to merge!

P.S. Additionally, if you could include essentially the notes you wrote above into the documentation, that would be great. I think I will reorganize the user guide a bit at some point soon so don't worry so much about where it goes, so long as it's pasted there somewhere... thanks!

pmlugato (approved):

lucalavezzo:
Thanks a lot Hasan. I have done some basic testing of both the old retriever and your new Hybrid one, everything seems to work, so I will merge now.

Look forward to testing the performance.

Issue #245, opened 2025-08-11
Clean up CLI


pmlugato:
With more interfaces and options being added all the time, CLI is getting quite long and not always consistent/clear -- needs to be reorganized.

ipausUchicago:
With regards to this, I think it would be useful if we had logging functionality in this document instead of using the _print_msg function, although im not sure if its necessary in some way. We could also benefit from having some more abstraction in loading all of the various services, perhaps we have different functions for each. It might be useful to have all of the config options and various relevant passwords/secrets all dealt with in one place and then have other functions to use them as necessary as we load the services being used.

PR #273, opened 2025-09-02
Made a config option in the cli to allow for the printing of the config file loaded into the a2rchi environment

ipausUchicago:


a2rchi/cli/cli_main.py
```diff
@@ -360,19 +363,22 @@ def create(
...
+            _print_msg("Preparing Grader Service")
         compose_template_vars["grader_image"] = f"grader-{name}"
         compose_template_vars["grader_tag"] = tag
 
         compose_template_vars["grader_volume_name"] = f"a2rchi-grader-{name}"
-        _create_volume(compose_template_vars["grader_volume_name"], podman=use_podman)
+        if not dry:
+          _create_volume(compose_template_vars["grader_volume_name"], podman=use_podman)
 
         _prepare_secret(a2rchi_name_dir, "admin_password", locations_of_secrets)
```
lucalavezzo, line 375: do we want to prepare secrets in --dry mode? Similar as above. It's sort of up to what we decide we mean by "dry" mode. What is most useful? What is the use case for this? If it's just to dump the config, I would think we don't need to preare secrets, make dirs, etc. But maybe you intended to want to check that all those things are being done correctly, without creating volumes. In any case, let's decide, and describe it more clearly in the parser.

a2rchi/cli/cli_main.py
```diff
@@ -360,19 +363,22 @@ def create(
...
 
         # prepare grader app logins file (users.csv)
         users_csv_dir = a2rchi_config['interfaces']['grader_app']['local_users_csv_dir']
         users_csv_path = os.path.expanduser(os.path.join(users_csv_dir, "users.csv"))
-        _print_msg(f"Preparing users.csv from {users_csv_path}")
+        if not dry:
+            _print_msg(f"Preparing users.csv from {users_csv_path}")
         if not os.path.isfile(users_csv_path):
             raise FileNotFoundError(f"users.csv file not found in directory {users_csv_dir}")
         target_users_csv_path = os.path.join(a2rchi_name_dir, "users.csv")
```
lucalavezzo, line 384: same comments as above: do we want to create these files in --dry mode?

lucalavezzo:
Merging for now, but we should move discussion to #245

PR #277, opened 2025-09-04
Fix cli_main: print -> logging

axion66:
Fix issue #273

Replaced _print_msg with logging.
Considered adding custom verbosity, but think t’s better to remove the custom verbose config and use the base config file from setup_logging.

Changed:

-config = config_template.render(verbosity=verbosity, **a2rchi_config)

to:

-config = config_template.render(**a2rchi_config)

### DIFF
--- a/a2rchi/cli/cli_main.py
+++ b/a2rchi/cli/cli_main.py
@@ -13,6 +13,11 @@ import time
 import shlex
 import threading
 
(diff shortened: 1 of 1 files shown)


pmlugato:
Closed as issue already resolved from merged PR, https://github.com/mit-submit/A2rchi/pull/265

PR #265, opened 2025-08-20
Re-organization of A2rchi, Workflows, ChainWrapper; Prompt formatting and abstraction; CLI re-rewrite; Configuration Handling; Other misc. improvements

lucalavezzo:
### Class structure

This PR seeks to re-organize the classes that handle the chains. The new structure is:

- `A2rchi`: this global class is what is called by the interfaces, updates the vectorstore connection, and invokes a given Pipeline selected via the configuration.
- `Pipeline`: sequences of `langchain`'s chains, retrievers, etc. (previously `BaseChain`s). Each A2rchi instance would pick a `Pipeline` to operate. e.g. The basic Q/A is QAPipeline, which defines the set of operations that are called: first a condensing step of the history, then the retrieval of documents, and finally the answer with the documents in the context.
- `ChainWrapper`: a lightweight, generic class to wrap `langchain`'s chains, to harmonize them with our prompts and inputs.

### Prompt abstraction

Further, prompts are abstracted in the following ways:

- The order of the input variables is now completely flexible.
- Which input variables are used is flexible within each `Pipeline`; each `Pipeline` can define a set of required input variables.
- Prompts are fully defined via the config. Any prompt can be used in any`Pipeline`.
- Required/optional prompts for a given `Pipeline` configured via the config

### CLI

Complete re-write. Pietro might want to add more here: TODO

### Config updating in A2rchi

The config file is loaded in many parts of the codebase, previously statically, at script init, and not again.
For all parts concerning the `A2rchi` class, the config is loaded only through the class itself. It can be updated with `A2rchi.update()`. This allows tighter control over what config we are currently using, with the idea we can set up a system to switch between configs now.
Global and interfaces configs are still considered static, i.e. loaded at script init time, and not again.

### Other

- `Cleo` -> `Redmine` everywhere
- `PromptLogger` now in a new utilities script
- `TokenLimiter` bug fix for the documents reduction, and expanded to cover any given input variable, which for now is left as unprunable. Whether an input is a list of documents is automatically determined.
- `{context}` -> `{retriever_output}` and `{condensed_question}` -> `{condensed_output}`. The idea being that soon we can move to a more generic MODEL_output.
- Chat shows top 5 links above a certain threshold; removes duplicate links

Issue #249, opened 2025-08-12
Add capability for multiple prompts in a single deployment


juanpablosalas:
In a particular deployment, A2rchi may make use of the same input documents but different main prompts may be utilized.

Instead of a single path for a MAIN_PROMPT in the config.yaml [1], multiple paths must be passed separated by a comma. They will be saved as prompts in [chains/prompts.py](https://github.com/mit-submit/A2rchi/blob/main/a2rchi/chains/prompts.py) using "QA_{k}" where the variable k is used as a counter. By default, the prompt loaded in [chains/chain.py](https://github.com/mit-submit/A2rchi/blob/main/a2rchi/chains/chain.py) will be QA_0. I have yet to understand fully how to set up the chat interface to change the prompt before inputting questions.

[1] https://mit-submit.github.io/A2rchi/

lucalavezzo:
Hey Juan, I am working on abstracting the prompts (the issue is here but I haven't documented anything yet: https://github.com/mit-submit/A2rchi/issues/231). I don't know if you're working on this already, but if you are we should avoid overlapping work :) I can make a draft PR sooner to show you what I had in mind, but I would see the abstraction as a necessary first step to adding multiple prompts.

PR #283, opened 2025-09-10
Capability for multiple configurations

juanpablosalas:
Based on #249 , we decided it would be better to have multiple config files loaded and validated and be able to switch them on the spot. This is fairly possible due to the big reorganization of the pipeline #265. 

Right now, this is an initial draft PR not ready for review as it depends heavily on the reorganization PR being finished and approved.


pmlugato (approved):
Looks good! Thanks again for the further back and forth on slack for remaining fixes -- will merge now!

PR #290, opened 2025-09-24
Added Benchmarking functionality

ipausUchicago:
A2rchi now has benchmarking functionality, it can easily be run by doing a2rchi evaluate and specifying either a list or a directory of configs to run it on. Feel free to give it a shot. Included in this pr is also a plots folder which can be used to later examine the data produced by benchmarking. 

**IMPORTANT - before running the script, ensure that the output directory both exists and is set correctly in all of the configs that you want to test on


ipausUchicago:
unclear why this is happening but for some reason the test is timing out here but its still launching locally on my end (although possibly it takes longer than before to launch).

pmlugato (approved):
I will go ahead and approve this, thanks for setting this up, Ivan, it's a really nice start and was much needed! Will merge now and follow up with a few small changes.

PR #294, opened 2025-10-03
Reorganization and sources

pmlugato:
- Relocates the codebase into a src/ package tree and updates CLI/runtime imports accordingly.
- Refactors the CLI around source registry, mirroring service registry, and rewrote some of the managers and helpers for better functionality/robustness as well as clarity
- Rebuilds the data manager into dedicated persistence, scraper, ticket, and vectorstore components; better uniformity across scrapers, tickets.
- Documentation updates coming...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,6 @@ dependencies = [
     "podman-compose==1.4.0",
     "python-dotenv>=0.19.0",
     "mkdocs==1.6.1",
-    "mkdocs-material==9.6.21",
     "ipykernel==6.30.1",
     "ipython==8.18.0",
     "matplotlib==3.9.4",


configs/cern.list
lucalavezzo, line 1: this is in my opinion not a good example of a set of source documents, as they require CERN auth, which the average user will not have

src/cli/utils/helpers.py
```diff
@@ -42,7 +44,9 @@ def parse_sources_option(ctx, param, value):
     if not value:
         return []
     
-    available_sources = ['jira', 'redmine']
+    available_sources = [
+        name for name in source_registry.names() if name != 'links'
```
lucalavezzo, line 48: why are links treated differently?

src/data_manager/collectors/persistence.py
```diff
@@ -0,0 +1,136 @@
...
+class PersistenceService:
+    """Shared filesystem persistence for collected resources."""
+
+    def __init__(self, data_path: Path | str) -> None:
+        self.data_path = Path(data_path)
+        self.sources_path = self.data_path / "sources.yml"
+        self.websites_dir = self.data_path / "websites"
+        self.git_dir = self.data_path / "git"
+        self.tickets_dir = self.data_path / "tickets"
+        self.tickets_index_path = self.data_path / "tickets.yml"
```
lucalavezzo, line 27: I am having a hard time understanding: are sources.yml and tickets.yml the two different databases (one for links, one for tickets) for the mapping of useful metadata for each document?

lucalavezzo:
Thanks a lot for the impressive work Pietro. I put a couple of comments I would like to resolve before merging; I think they're only clarifications. A couple more things I think we might as well address in this PR:

- [ ] get rid of `configs/`, leaving only an example one, a benchmarking example one, and the PR workflows one
- [ ] similarly, get rid of all `prompts/` except for example one(s)
- [ ] I don't think `plots/` follow standard naming practices. Maybe it can be `scripts/`? So like `scripts/benchmarking/`

src/cli/templates/base-config.yaml
```diff
@@ -134,6 +132,59 @@ data_manager:
   bm25:
     k1: {{ data_manager.bm25.k1 | default(0.5, true) }}
     b: {{ data_manager.bm25.b | default(0.75, true) }}
+  sources:
+    links:
+      enabled: {{ data_manager.sources.links.enabled | default(true, true) }}
```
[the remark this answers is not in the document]
lucalavezzo: (ditto for the other sources)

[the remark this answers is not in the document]
pmlugato: `enabled` is what the sources look for later, you are meant to pass them (sso, jira, etc.) via the CLI with `--sources`, which updates config accordingly. `links` in particular are automatically enabled, which is dumb actually so will change.

general note, I agree with moving git away from scraping/link logic like mentioned in that issue we made, but since code is set up like this for now, can be a separate PR to move it away, I think. Then, `sso` could live under `data_manager: sources: links`, and git remain a separate source like jira or redmine...

src/cli/utils/helpers.py
pmlugato: automatically enabled but changing -- addressed in other comment

src/cli/templates/base-config.yaml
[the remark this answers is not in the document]
lucalavezzo: but we tell users to go set up the configuration file by hand, so they might think that setting `enabled` is enough, rather than passing with `--sources`. Would this break anything? (to set `enabled` without `--sources`)

PR #312, opened 2025-10-15
Generalize metadata and cataloguing of sources

lucalavezzo:
**Persistent storage**
- the persistent storage mechanism now allows for the storing of resource metadata in separate yaml files to the original content
- metadata is sanitized before writing out

**Resource classes**
- the metadata for each type of resource is defined at the Resource class level
- also at the Resource level, we now define: file hash, file name, content to be written out
- from the point of view of the persistent storage and the vector database, all sources are not standardized, allowing for seamless integration of new sources / collectors

**Catalogues**
- the catalogue now stores `file hash : file path`
- added new catalogue for metadata with `file hash : metadata path`, with the `file hash` being the same as the actual file of which the metadata belongs, to provide a mapping between the two

**Vector database**
- all metadata is now written in the vector database in the same Document object as the content

**Metadata**
- generalized class for ResourceMetadata added
- metadata is enforced to have a `display_name` field: this is what we can use in the apps to point to the original source

**Chat app**
- the chat app now displays the `display_name` metadata field for all sources (not just web links), and, if a link if found in the metadata, will show it
- each source can now be `visible` or not, via the configuration: i.e., the chat app will not display what the source used if its type is not visible. This is useful, for example, to avoid showing SubMIT users which Redmine ticket provided information about this response

**Other**
- added an optional limit to the number of tickets to process in JIRA for testing purposes

### DIFF
--- a/src/interfaces/redmine_mailer_integration/redmine.py
+++ b/src/interfaces/redmine_mailer_integration/redmine.py
@@ -63,17 +63,10 @@ class RedmineAIWrapper:
         if num_retrieved_docs > 0:
             for k in range(num_retrieved_docs):
                 document = source_documents[k]
-                document_source_hash = document.metadata['source']
-                if '/' in document_source_hash and '.' in document_source_hash:
-                    document_source_hash = document_source_hash.split('/')[-1].split('.')[0]
-                link_k = "link not available"
-                if document_source_hash in sources:
-                    link_k = sources[document_source_hash]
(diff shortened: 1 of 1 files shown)


pmlugato (approved):
Approved, thanks a lot for the development -- and for the fast updates regarding the discussion in person! Merging now, looking forward to the upcoming benchmarking improvements to expand link checking to all sources.

PR #316, opened 2025-10-16
New GitHub workflow for tag release

lucalavezzo:
**Workflow for new tag release**
New workflow to make a tag will:
1. build images with the current Dockerfiles, push them to DockerHub as :tag
2. run smoke test with the new images
3. push the new images to DockerHub as :latest
4. update the pyproject version
5. update the docs version, and re-deploy the docs
6. make git tag
Each successive step fails if the previous one fails. We could in the future add more tests.

**Misc.**
- docs now suggest to switch to the latest stable and available tag
- added some scripts and README's for common dev tasks like creating, pushing images, etc.
- changed the default GPU model to a non-reasoning model, since we don't sanitize or properly display those outputs yet
- wording changes to project README

### DIFF
--- a/.github/workflows/test-and-build-tag.yml
+++ b/.github/workflows/test-and-build-tag.yml
@@ -105,6 +105,24 @@ jobs:
           wait-url: http://localhost:7861/api/health
           base-url: http://localhost:7861
 
+      - name: Commit Dockerfile base image updates
+        if: success()
+        env:
+          TARGET_REF: ${{ inputs.ref || 'main' }}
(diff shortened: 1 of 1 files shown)


pmlugato:
Thanks for adding this @lucalavezzo , and for following up discussions locally. Everything looks good to me, I will approve, and for reference, below is what we said you will have in the following PR:

1. a dev image which will always be up to date with current version of main, for when changes to the base image are made but no tag is yet released.
2. with a new PR, if the base image (the requirements) is changed, the test needs to build the base image on the fly and use that to test, so handling for this

Issue #310, opened 2025-10-13
match the uploader interface to the chatbot interface (also for other services)


juanpablosalas:
In order to preserve uniformity, all the interfaces (such as the document uploader) should match the main (chatbot) UI. Things like colors and fonts, should be relatively easy to change we just need to find the time as it's not main priority.

PR #322, opened 2025-10-22
Initial version of document uploader UI

juanpablosalas:
This is an initial version of having a user interface to view the input documents as suggested in #310.

I also included a couple of changes such as:
- Having a title saved in metadata of git and sso scraped resources
- Updates to how the document uploader was listing documents
-  Improvements in the `delete_url` function of the document uploader

So far, this only works to preview git scraped documentation and scraped urls (whether sso or not, whether uploaded "manually" or via the input lists) but in the future the idea is to extend it to preview manually uploaded files (i.e. pdf, txt, etc).

Also, it could be better if we could create the service accounts directly from the UI instead of how it's done now (running a shell within the container and running the python script)

The list of documents is named by the file hashes as this was the most standard way to do it. In the future, we could rethink this into listing them as a file tree (?) in order to show the relationships between the link crawling (i.e. a single sso link generates tens of documents so it would be important to know that this all came from the same "source"). B

Screenshot 2025-10-22 at 11 25 22 AM

As always, I'm open to suggestions and comments! :)

### DIFF
--- a/.github/actions/run-smoke/action.yml
+++ b/.github/actions/run-smoke/action.yml
@@ -76,9 +76,7 @@ runs:
         ENV_FILE: ${{ inputs.env-file }}
       run: |
         set -euo pipefail
-        echo "PG_PASSWORD=$(openssl rand -base64 32)" > "${ENV_FILE}"
-        echo "UPLOADER_SALT=$(openssl rand -base64 32)" >> "${ENV_FILE}"
-        echo "FLASK_UPLOADER_APP_SECRET_KEY=$(openssl rand -base64 32)" >> "${ENV_FILE}"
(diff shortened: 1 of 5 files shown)


pmlugato:
Thanks for this PR @juanpablosalas, and for the review and discussion it prompted on slack :) Summarizing below:

1. no longer dedicated service, but another chat endpoint
2. visualization of all sources (not just uploaded manually) organized by source type
3. option to add/delete, document by document, or by metadata filter.

pmlugato:
@juanpablosalas Thanks for all of the work on this, it's really nice, and for following up on the discussion on slack with @lucalavezzo and I. I saw you opened issues on a couple things that should follow this PR, so thanks for that too. I will approve now and merge!

PR #324, opened 2025-10-23
Remove block_text and redundant <p> statements from chat_app/app.py

daniel0tgc:
This PR fixes a bug where LLM-generated Markdown (tables, lists, horizontal lines) was rendering as raw text instead of HTML.

Three main changes were done:
Removed block_text from AnswerRenderer: This custom method was incorrectly wrapping all non-code text in <p> tags, preventing the mistune parser from recognizing tables or lists.

Enables the table plugin: The format_code_in_text method is updated to pass plugins=['table'] when creating the Markdown parser, explicitly enabling table rendering.

Removes redundant <p> wrapper: The final output in ChatWrapper.__call__ is no longer wrapped in <p> tags, which was creating invalid HTML (e.g., <p><table>...</table></p>).


src/interfaces/chat_app/app.py
```diff
@@ -549,8 +549,10 @@ def __call__(self, message: List[str], conversation_id: int|None, is_refresh: bo
             logger.info(f"Number of queries is: {self.number_of_queries}")
 
             # display answer
-            output = "<p>" + self.format_code_in_text(result["answer"]) + "</p>"
 
+            logger.info(f"--- DEBUG: Raw answer from A2rchi: {result['answer'][:150]}... ---")
+            output = self.format_code_in_text(result["answer"])
+            logger.info(f"--- DEBUG: Final rendered output: {output[:150]}... ---")
```
pmlugato: I think you can remove the two logger statements; other logging already prints output and for debugging unformatted code this can be added back manually

PR #327, opened 2025-10-28
Agents and tools

lucalavezzo:
**Reorganization**
- pipelines now are either 'classic pipelines' (based on langchain-classic's Chains), or agents. The user can choose either from the configuration
- pipelines have a standardized output `PipelineOutput`
- models are now in a directory, and each one lives in its own file
- retrievers are now in a directory, and each one lives in its own file

**Agents**
- added langchain's agents as possible pipelines
- wrote a first version of an agent, CMSCompOpsAgent

**Tools**
- added tools for agents to: query the vectorstore, read local files from the catalogue, read metadata from the catalogue
- added DocumentCollector to return to the pipeline the documents returned from the various internal steps of the agents

**Catalogue handling**
- added a CatalogueService to easily interact with the catalogue: read documents from the hash, read metadata from the hash, get file paths from the hash, etc.

**Chat app**
- [in progress] added support for agent streaming

### DIFF
--- a/src/data_manager/collectors/persistence.py
+++ b/src/data_manager/collectors/persistence.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from pathlib import Path
-from typing import Dict, TYPE_CHECKING, Union, Any, List
+from typing import Dict, TYPE_CHECKING, Union, Any
 
 import yaml
 
@@ -21,12 +21,8 @@ class PersistenceService:
     def __init__(self, data_path: Path | str) -> None:
         self.data_path = Path(data_path)
 
-        self._index: Dict[str, str] = CatalogService.load_index(self.data_path)
+        self.catalog = CatalogService(self.data_path)
         self._index_dirty = False
-
-        self._metadata_index: Dict[str, str] = CatalogService.load_index(
-            self.data_path, filename="metadata_index.yaml"
-        )
         self._metadata_index_dirty = False
 
     def persist_resource(self, resource: "BaseResource", target_dir: Path) -> Path:
@@ -52,7 +48,7 @@ class PersistenceService:
                 metadata_relative_path = str(metadata_path)
 
             resource_hash = resource.get_hash()
-            self._metadata_index[resource_hash] = metadata_relative_path
+            self.catalog.metadata_index[resource_hash] = metadata_relative_path
             self._metadata_index_dirty = True
 
         try:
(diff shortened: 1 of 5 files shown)


src/cli/managers/templates_manager.py
```diff
@@ -13,6 +13,51 @@
...
+    robust than assuming a fixed number of parent hops from __file__.
+    """
+    markers = ("pyproject.toml", "LICENSE", ".git")
+    for m in markers:
+        if (path / m).exists():
+            return True
+    return False
+
+
+def find_repo_root() -> Path:
```
pmlugato: can you make an issue to change this such that the `src/` is included in the a2rchi package? I think @ipausUchicago was implementing this at some point for this same reason, to not have to be at the repo root to deploy, so we can assign Ivan to this.

src/a2rchi/utils/vectorstore_connector.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+        self.collection_name = dm_config["collection_name"] + "_with_" + embedding_name
+        self.use_HTTP_chromadb_client = chroma_config["use_HTTP_chromadb_client"]
+        self.chromadb_host = chroma_config["chromadb_host"]
+        self.chromadb_port = chroma_config["chromadb_port"]
+        self.local_vstore_path = chroma_config["local_vstore_path"]
+
+        logger.info(f"Vectorstore connection initialized with collection: {self.collection_name}")
+
+    def _update_vectorstore(self):
```
pmlugato: change name to not make it seem like the contents of the vector database are being updated (which was once the case but now this is just connecting now)

src/a2rchi/pipelines/agents/utils/document_collector.py
```diff
@@ -0,0 +1,79 @@
+"""Utilities for aggregating documents and notes during agent runs."""
+
+from __future__ import annotations
+
+from typing import Iterable, List, Sequence, Tuple
+
+from langchain_core.documents import Document
+
+
+class DocumentCollector:
```
pmlugato: Can we use a different name for this :)

pmlugato:
@lucalavezzo This is really great work, thanks for doing the heavy lifting of getting the infrastructure together for agents going. I left some comments, I also will add @hassan11196 as a reviewer, as he will test this setup with the CompOps deployment living at CERN, so if there are things you notice, Hassan, please add comments or suggestions to the PR (immediate changes or things to update/change later for which we can make issues).

pmlugato (approved):

PR #328, opened 2025-10-29
Add crawling functionality to the base webscraper

ipausUchicago:
Extended the functionality of the base webscraper, made it more simple to specify the depth to be crawled to on each webpage, you can now optionally pass in a comma separated depth at the end of the line where the url is specified in the file. by default this is set to one if not specified. Additionally Refactored the SSo scraper to make a base selenium scraper class. the idea being that it leaves it open for future support with javascript loaded webpages by writing another scraper on top of it.

### DIFF
--- a/src/cli/templates/base-config.yaml
+++ b/src/cli/templates/base-config.yaml
@@ -148,8 +148,8 @@ data_manager:
       bm25_k1: {{ data_manager.retrievers.hybrid_retriever.bm25_k1 | default(0.5, true) }}
       bm25_b: {{ data_manager.retrievers.hybrid_retriever.bm25_b | default(0.75, true) }}
   sources:
-    base_source_depth: {{ datamanager.sources.base_source_depth | default(1, true) }}
     links:
+      base_source_depth: {{ datamanager.sources.links.base_source_depth | default(1, true) }}
       enabled: {{ data_manager.sources.links.enabled | default(true, true) }}
       visible: {{ data_manager.sources.links.visible | default(true, true) }}
       input_lists:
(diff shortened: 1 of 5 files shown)


src/data_manager/collectors/scrapers/integrations/sso_scraper.py
```diff
@@ -227,6 +227,9 @@ def crawl(self, start_url):
         logger.info(f"Crawling complete. Visited {pages_visited} pages.")
         return list(self.page_data)
 
+    def _():
```
pmlugato: What is this for?

src/data_manager/collectors/scrapers/scraper.py
```diff
@@ -1,57 +1,161 @@
...
+                        level_links.append(link)
+
+            except requests.exceptions.HTTPError as e: 
+                if e.response.status_code in (401, 403):  # if the resonse was an auth error reauthenticate with the authenticator
+                    if not self.authenticator is None:
+                        cookies = self.authenticator.authenticate(current_url)
+                        try:
+                            # note that we could expect two errors, 1) antoher http error 2) selenium wasnt able to get auth so cookies is None
+                            for cookie in cookies: 
+                                s.cookies.set_cookie(cookie['name'], cookie['vaule'])
```
pmlugato: typo:`vaule` --> `value`

pmlugato:
@ipausUchicago thanks for the updates. So, I assume that the selenium-based crawler is more robust with getting content from different types of web pages as opposed to just using the requests package, but at the cost of being slower, is this right? Maybe @hassan11196 also has some ideas about this having set up the selenium-based version, e.g., if there are CERN pages that requests does not correctly parse. I suppose we can keep both for now, like you've coded up. But we should probably think a bit more about how to handle this. It doesn't make much sense to me to decide whether to use requests- vs selenium-based scraping depending on whether we need to authenticate or not.

TLDR; we should be careful to make sure that requests is indeed correctly getting the content from webpages!

In any case, second order problem, it's great to see crawling for non-sso links now, so thanks a lot, Ivan!

src/cli/templates/base-config.yaml
```diff
@@ -154,6 +155,7 @@ data_manager:
     sso:
```
lucalavezzo: since `sso` is now just a setting of `links`, can we move these parameters under `links`? Also, can we remove `sso` from `sources` at CLI?

src/data_manager/collectors/scrapers/integrations/sso_scraper.py
```diff
@@ -264,6 +264,25 @@ def authenticate_and_navigate(self, url):
         except Exception as e:
             logger.warning(f"Error during authentication: {e}")
             return None
+
+    def authenticate(self, url):
```
lucalavezzo: if I understand correctly, the SSOScraper now provides the WebScraper with the authentication, and nothing else, right? Since now the WebScraper does everything the SSOScraper could do. Can we then remove the SSOScraper entirely, and just have a SSOAuth, or something of that sort? In short, there should only be one collector to scrape websites, the SSOAuth should just be a base class for the different specific auths like CERNSSOScraper. If you have a better proposal, let me know

lucalavezzo (approved):

lucalavezzo:
Thanks a lot @ipausUchicago, I'm making some small changes in the next PR, but this seems to work now!

PR #334, opened 2025-11-05
benchmarking upgrades; fix setup paths; parallelize chunking; agents hotfixes

lucalavezzo:
**Benchmarking**:
- Added script from @haozturk for viewing results of benchmarking run
- Added full message chain in the benchmarking output and to the script
- Misc. clean up of the benchmarking service

**Fix setup, installation paths**
- now one can run `a2rchi` commands from anywhere, and the PR preview works

**Parallelizing chunking**:
- parallelize the chunking process before adding to vectostore. Speeds up the data ingestion significantly

**Other**:
- misc. hotfixes from previous PR adding agents

### DIFF
--- a/src/bin/service_benchmark.py
+++ b/src/bin/service_benchmark.py
@@ -382,7 +382,8 @@ class Benchmarker:
             ragas_input = []
 
             # SOUCES mode: sources accuracy
-            source_accuracy = 0.0 
+            relative_source_accuracy = 0.0 
+            source_accuracy = 0.0
 
             for question_item in self.queries_to_answers:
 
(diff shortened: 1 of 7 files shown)


src/bin/service_benchmark.py
```diff
@@ -338,28 +358,41 @@ def get_ragas_results(self, data, to_add):
 
         return res
 
+
     def run(self):
         modes_being_run = set(self.benchmarking_configs['modes'])
```
juanpablosalas: I was checking the service builder and I don't think we're checking what the user is putting as modes in the configuration. Also, the default modes are set to "LINKS" and "RAGAS" but I think you refer to the first one as "SOURCES", or is the first a mode that has not been implemented yet?

lucalavezzo: Thanks for looking over this! Let me try to address your comments.

> I don't think we're checking what the user is putting as modes in the configuration.

yea, I guess we don't check this anywhere -- were you suggesting we do? if so, where?

> Also, the default modes are set to "LINKS" and "RAGAS"

is this in the base-config? wherever that is, let's change it.

> or is the first a mode that has not been implemented yet?

indeed, I renamed LINKS to SOURCES, when I expanded the usage to include all types of sources, thought I got rid of all vestiges, but if you find any left, I can remove them!

PR #337, opened 2025-11-20
Fix list out of index bug

haozturk:
Fixes:
```
(2025-11-19 15:26:38,902) [src.interfaces.chat_app.app] ERROR: Failed to produce response: list index out of range
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/src/interfaces/chat_app/app.py", line 570, in __call__
    top_sources = self.get_top_sources(documents, scores)
  File "/usr/local/lib/python3.10/site-packages/src/interfaces/chat_app/app.py", line 170, in get_top_sources
    documents = [documents[i] for i in sorted_indices]
  File "/usr/local/lib/python3.10/site-packages/src/interfaces/chat_app/app.py", line 170, in <listcomp>
    documents = [documents[i] for i in sorted_indices]
IndexError: list index out of range
```
I think this happens because documents are pruned by token_limiter while scores aren't. e.g.
```
score length: 8 - scores: [-1.0, -1.0, -1.0, -1.0, -1.0, -1.0, -1.0, -1.0]
document length: 5 - documents: [TRUNCATED]
```
The score-document logic should be revisited altogether given that it's only valid for semantic retrieval. This PR is a workaround until that so that the version on the main branch works.

### DIFF
--- a/src/interfaces/chat_app/app.py
+++ b/src/interfaces/chat_app/app.py
@@ -164,10 +164,27 @@ class ChatWrapper:
         """
         Build a list of top reference entries (link or ticket id).
         """
+        # TODO: It's better to ensure in token limiter that scores and documents are pruned together and have the same length
         if scores:
-            sorted_indices = np.argsort(scores)
-            scores = [scores[i] for i in sorted_indices]
-            documents = [documents[i] for i in sorted_indices]
+            # Ensure scores and documents have the same length
+            min_length = min(len(scores), len(documents))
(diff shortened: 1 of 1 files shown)


lucalavezzo:
I don't think we need this at all @pmlugato @haozturk, or at least, we don't understand what's happening to cause this issue. This is fine as bandaid fix for now, but I would just ask to add a logger.warning() before we merge, just so we make it clear that something fishy is going on.

Issue #320, opened 2025-10-21
Revisit handling of scores


haozturk:
Currently pipelines.py expects a docs, score tuple from retrievers. Although, semantic retriever returns a score by default, BM25 doesn't. We need to find a way to make it work for all retrievers.

PR #340, opened 2025-11-21
Address issues in CompOps deployment

lucalavezzo:
**Retrievers**
- Split BM25 to be its own retriever
- Set up CMSCompOpsAgent to have Semantic and BM25 as different Retrievers, in the hope it will form more apt queries for each
- Remove unnecessary retriever calls in HybridRetriever
- Re-organized configuration yaml file for retrievers

**Chat app**
- Don't show missing scores
- Small improvements for debugging purposes of source visibility

### DIFF
--- a/src/interfaces/chat_app/app.py
+++ b/src/interfaces/chat_app/app.py
@@ -66,14 +66,10 @@ class AnswerRenderer(mt.HTMLRenderer):
             "mathematica" : MathematicaLexer,
             "matlab": MatlabLexer
         }
-    
+
     def __init__(self):
         self.config = load_config()
         super().__init__()
-
-    def block_text(self,text):
-         #Handle blocks of text (the negatives of blocks of code) and sets them in paragraphs
-         return f"""<p>{text}</p>"""
 
     def block_code(self, code, info=None):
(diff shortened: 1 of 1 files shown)


lucalavezzo:
FYI @haozturk I don't know if this addresses #320, I don't even think that the agent stores the scores properly at the moment. Could you try to see if you still see the error?

PS the DocumentMemory, which is what we use in the agent to "remember" documents that are fetched during internal steps, doesn't handle scores at all.

src/a2rchi/pipelines/agents/cms_comp_ops_agent.py
```diff
@@ -119,38 +119,56 @@ def _prepare_agent_inputs(self, **kwargs) -> Dict[str, Any]:
                 memory.note(f"Latest user message: {snippet}")
         return {"messages": history_messages}
 
-    def _update_vector_retriever(self, vectorstore: Any) -> None:
+    def _update_vector_retrievers(self, vectorstore: Any) -> None:
```
pmlugato: BM25 not vector search, but just a naming thing... logic is otherwise good

PR #353, opened 2025-12-01
merged both versions of html report

juanpablosalas:
This PR combines the view_results scripts for the benchmarking with @haozturk. It also automates the generation of the HTML report. As this is the main format we have been using for sharing benchmarking results, I thought it might be worth to unify.

[1] https://gitlab.cern.ch/a2rchi/cms-compops/-/blob/haozturk-submit76/scripts/compare_benchmark_outputs.py?ref_type=heads

### DIFF
--- a/requirements/requirements-base.txt
+++ b/requirements/requirements-base.txt
@@ -18,7 +18,7 @@ humanfriendly==10.0
 langgraph==1.0.1
 langchain-mcp-adapters==0.1.11
 langchain==1.0.2
-langchain-ollama==1.0.0
+langchain_ollama==1.0.0
 loguru==0.7.2
 lz4==4.3.2
 mistune==3.0.2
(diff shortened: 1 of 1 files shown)


requirements/requirements-base.txt
```diff
@@ -42,7 +42,7 @@ zstandard==0.25.0
 langchain-openai==1.0.1
 langchain-chroma==1.0.0
 langchain-community==0.4
-langchain-core==1.0.0
+langchain-core==1.0.2
```
lucalavezzo: I'm not against this, but why is it needed in this PR?

juanpablosalas: Agree, this was just to run RAGAS to generate the full report, however I removed it from the PR as it's more relevant to when we take care of #363

PR #364, opened 2025-12-05
Persistent feedback option, more robust conversation management, some UI, bug fixes...

pmlugato:
*Feedback*
- Written feedback option present for every message, can submit however many comments which are tracked and shown to user

*Bug fixes*
- clearing a conversation removes it locally but remains in postgres -- can consider adding an option to prune from database too but not needed for now
- likes and dislikes can only be recorded once per message (multiple clicks registered as multiple likes before :)
- refresh button behaves now, so if clicked, nothing is removed from database (e.g., previous message's negative feedback) and only most recent is considered in the history. also rendered correctly
- grafana gets timing information now, SQL query was outdated

*UI + Grafana*
- Multiple conversations for given user treated as such in grafana now
- config switching option a pop up menu from icon with some prettying up. also added the `services.chat_app.trained_on` field as description and as such not requiring it to be constant across the multiple configs
- feedback charts in grafana are with respect to different configs

*Misc*
- some SQL helpers for some features mentioned above
- localStorage improvements to keep clientIDs, cleared convos, across redeployments, resetting correctly when a **new** deployment is made, i.e. new --name or volumes were deleted.

### DIFF
--- a/src/cli/templates/grafana/a2rchi-default-dashboard.json
+++ b/src/cli/templates/grafana/a2rchi-default-dashboard.json
@@ -248,12 +248,12 @@
           },
           "mappings": [
 {% for palette in feedback_palette %}
-            {% set base = loop.index0 * 4 %}
+            {% set base = loop.index0 * 2 %}
             {
               "options": {
-                "pattern": "like_with_config_{{ palette.name }}",
+                "pattern": "like_{{ palette.name }}",
                 "result": {
-                  "color": "{{ palette.like_with_config }}",
+                  "color": "{{ palette.like }}",
                   "index": {{ base }}
                 }
               },
@@ -261,30 +261,10 @@
             },
             {
               "options": {
-                "pattern": "like_no_context_{{ palette.name }}",
+                "pattern": "dislike_{{ palette.name }}",
                 "result": {
-                  "color": "{{ palette.like_no_context }}",
+                  "color": "{{ palette.dislike }}",
                   "index": {{ base + 1 }}
-                }
-              },
-              "type": "regex"
-            },
-            {
-              "options": {
-                "pattern": "dislike_with_config_{{ palette.name }}",
-                "result": {
-                  "color": "{{ palette.dislike_with_config }}",
-                  "index": {{ base + 2 }}
-                }
-              },
-              "type": "regex"
-            },
-            {
-              "options": {
-                "pattern": "dislike_no_context_{{ palette.name }}",
-                "result": {
-                  "color": "{{ palette.dislike_no_context }}",
-                  "index": {{ base + 3 }}
                 }
               },
               "type": "regex"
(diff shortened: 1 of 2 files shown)


src/interfaces/chat_app/utils.py
```diff
@@ -0,0 +1,24 @@
...
+    Works for both simple history tuples and extended rows with metadata.
+    """
+    if not history_rows:
+        return history_rows
+
+    collapsed = []
+    assistant_run = []
+    for row in history_rows:
+        sender = row[sender_index]
+        if sender == "A2rchi":
```
lucalavezzo: we should formalize these senders in the chat app, such that we have like A2RCHI_SENDER defined globally. Else one minute change one place can cause a whole series of headaches.

src/interfaces/chat_app/app.py
```diff
@@ -1062,6 +1166,50 @@ def dislike(self):
             if self.chat.conn is not None:
                 self.chat.conn.close()
 
+    def text_feedback(self):
```
lucalavezzo: maybe to reduce load in this app.py, we can make a utilities file with all the feedback-related functions

src/cli/templates/base-config.yaml
```diff
@@ -67,7 +67,7 @@ services:
     host: {{ 'localhost' if host_mode else (utils.postgres.host | default('postgres', true)) }}
   chat_app:
     pipeline: {{ services.chat_app.pipeline | default("QAPipeline", true) }}
-    trained_on: {{ services.chat_app.trained_on | default("", true) }}
+    trained_on: {{ services.chat_app.trained_on | default("No description provided.", true) }}
```
lucalavezzo: change this to landing_page_text or something? could even think about allowing for custom html to be put there instead.

src/interfaces/chat_app/utils.py
pmlugato: can put it as a constant at the top of the file, sure yeah

src/cli/templates/base-config.yaml
pmlugato: can put `agent_description` under `a2rchi` field. in theory data_manager can change too but ok if you change data_manager it's not crazy to just adjust the description there...

src/interfaces/chat_app/app.py
```diff
@@ -678,6 +772,7 @@ def __call__(self, message: List[str], conversation_id: int|None, client_id: str
                 is_refresh
             )
             timestamps['insert_convo_ts'] = datetime.now()
+            history.append(("A2rchi", result["answer"]))
```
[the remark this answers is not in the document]
pmlugato: addressed yep, thanks

lucalavezzo (approved):
Thanks for the iteration. This looks good to me now.

PR #374, opened 2025-12-15
Included updated and created date for git docs

juanpablosalas:
As suggested by Jason, I included the last updated and created dates for files retrieved by the git scraper. This is done using the git python library and the first and last commit associated with each file.

### DIFF
--- a/src/data_manager/collectors/scrapers/integrations/git_scraper.py
+++ b/src/data_manager/collectors/scrapers/integrations/git_scraper.py
@@ -2,9 +2,10 @@ import re
 import shutil
 from pathlib import Path
 from typing import TYPE_CHECKING, List, Tuple
+from datetime import datetime
 
 from mkdocs.utils.yaml import yaml_load
-from git import Repo
+from git import Repo, Commit
 
 from src.data_manager.collectors.scrapers.scraped_resource import \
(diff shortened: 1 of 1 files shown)


src/data_manager/collectors/scrapers/integrations/git_scraper.py
```diff
@@ -80,20 +84,25 @@ def _harvest_repository(
...
             current_url = base_site_url + \
                 markdown_path.relative_to(docs_dir).with_suffix("").as_posix()
 
             logger.info(f"Indexing Git doc: {current_url}")
             text_content = markdown_path.read_text(encoding="utf-8")
+            created_at = self._get_creation_date(repo=repo, file_path=markdown_path)
+            last_updated_at = self._get_last_updated_date(repo=repo, file_path=markdown_path)
             resource = ScrapedResource(
                 url=current_url,
                 content=text_content,
```
lucalavezzo: We could think about putting the file name, date, repo, etc. in the text_content to make it easier for things like BM25 to find files via query. What do you think?

lucalavezzo (commented):
Thanks @juanpablosalas, minor comments, but then we can merge. I would like to pull this into #377 and build some things on top of it.

PR #376, opened 2025-12-17
Add client_id to conversation load request payload

hassan11196:
Fix for issue of not able to switch to other conversations


haozturk:
Thanks for the quick fix Hassan, LGTM - Tested locally and it works

PR #377, opened 2025-12-18
Upgrades to data ingestion

lucalavezzo:
- Data manager as its own service, container, volume
- Updated the document uploader UI, which is its own app now
- Agent tools ping the data manager app via appropriate endpoints
- Generalized @juanpablosalas's JIRA scheduler: all sources can be scheduled to ingest. Behavior of each can be tuned (e.g. JIRA ingests only tickets since last run) via the collector. These can be set in config, or via the data-manager UI.
- Added local files as a source: these are mounted to the container at start up time. At the moment, we don't support adding new local files, except through the browser upload
- Turned on all sources by default

### DIFF
--- a/src/data_manager/collectors/scrapers/integrations/sso_scraper.py
+++ b/src/data_manager/collectors/scrapers/integrations/sso_scraper.py
@@ -360,24 +360,6 @@ class SSOCollector:
         except Exception as exc:  # pragma: no cover - defensive catch
             logger.error(f"SSO scraping failed for {url}: {exc}")
             return []
-    
-    def get_auth(self, url: str) -> Dict | None:
-        if not self._enabled: 
-            logger.error("SSO is disabled or not configured")
-            return None
-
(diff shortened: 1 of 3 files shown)


src/bin/service_benchmark.py
```diff
@@ -150,20 +148,10 @@ def load_new_configuration(self):
...
 
-        # for now we just reset the datamanager entirely every time,
-        # in the future we should add support for hot config swapping so 
-        # documents dont need to be reinputted unnecessarily
-        if self.data_manager:
-            del self.data_manager
-        self.data_manager = DataManager()
-        self.data_manager.update_vectorstore()
-        self.previous_input_list = current_input_list
-
```
pmlugato: is this deletion just making the benchmarking run over multiple configs constrained to the same data? I agree with this, should some config be tested over different datasets it can be run manually, so just confirming

PR #408, opened 2026-01-22
Updated ci_test_build for consistency

Viphava280444:
Hello,
I noticed that the pr-review configuration had a port change, so I updated it in the ci_test_build workflow as well.

Thank you


pmlugato (approved):
thanks Ohm, good catch! Approved and merging.

PR #409, opened 2026-01-22
now collects tools from active servers and tells which servers are ac…

NamN8Tr:
…tive and inactive


pmlugato (approved):
looks good, thanks for quick follow up @NamN8Tr , merging

PR #411, opened 2026-01-24
Improved smoke test; add openspec project files

lucalavezzo:
_**Improved smoke test**_

Smoke test now:

- preflight checks that all databases are healthy
- checks that important tools are working
- launches CMSCompOpsAgent with tools, ollama, and asks a simple question

This should allow us to probe that the launch, data ingestion, database creation, and agents are all somewhat working.

_**openspec**_
We have started to use openspec https://github.com/Fission-AI/OpenSpec to better guide coding agents.

**_Bug fix in redmine postgres_**
Small issue was causing redmine to crash, should be fixed now but haven't been able to test since we don't have a nice way to do so except through a live redmine server.

### DIFF
--- a/docs/docs/developer_guide.md
+++ b/docs/docs/developer_guide.md
@@ -77,6 +77,22 @@ export METADATA_SEARCH_QUERY="ppc.mit.edu"
 export VECTORSTORE_QUERY="cms"
 ```
 
+## Postgres Usage Overview
+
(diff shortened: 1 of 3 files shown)


pmlugato (approved):
Looks good to me, I'm sure we will find bugs when certain things change but can continue to fix along the way like we have been doing with the old test... merging

PR #445, opened 2026-02-16
patch broken tools available in CMSCompOpsAgent

lucalavezzo:


pmlugato (approved):
thanks for bug fix, merging

PR #456, opened 2026-02-19
Add ELOG scraper and deployment for FNAL dCache ELOG using FNAL Ollama server

nhduongvn:
- Add ElogScraper to crawl ELOG logbooks (pagination, entry parsing, structured metadata extraction: tech, category, node, incident_date, etc.)
- Support explicit `elog-<url>` prefix in input lists for unambiguous ELOG URL detection alongside the existing heuristic (_is_elog_url)
- Update cms-comp-ops agent prompt with ELOG tool guidance: use tech: field for person queries, cite url metadata (not internal hashes), clarify that [N] are result indices not ELOG entry numbers, note 5-result limit
- Add examples/deployments/basic-ollama-fnal with config targeting ollama.fnal.gov and the FNAL dCache ELOG as a data source

Assisted by Claude Sonnet 4.6

### DIFF
--- a/examples/deployments/basic-ollama-fnal/condense.prompt
+++ b/examples/deployments/basic-ollama-fnal/condense.prompt
@@ -2,7 +2,7 @@
 # This is a very general prompt for condensing histories, so for base installs it will not need to be modified
 # 
 # All condensing prompts must have the following tags in them, which will be filled with the appropriate information:
-#      {chat_history}
+#      {history}
 #      {question}
 #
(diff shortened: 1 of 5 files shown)


juanpablosalas (approved):
The e-log scraping functionality is working as expected. As this was moved to the dev branch I think it's okay to leave the example deployment and agent as it is. However, if we later want to move this to the main branch, I would suggest removing the example deployment as it overlaps mostly with other deployments and the only relevant changes are highly specific.

PR #462, opened 2026-02-20
Benchmarking fixes; chat streaming fixes

lucalavezzo:


src/cli/templates/dockerfiles/Dockerfile-benchmarks
```diff
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1
-FROM docker.io/a2rchi/a2rchi-python-base:latest 
+FROM localhost/a2rchi/a2rchi-python-base:latest
```
pmlugato: why are all these switched? was it for testing the upgraded base reqs?

pmlugato (commented):
Based on a quick review, this looks great and will wrap up a few of the things we wanted to make sure to have fixed in the upcoming v2 release, thanks a lot @lucalavezzo . I will look a bit more, plus await @haozturk 's comments, hence holding off on the approval, but I've also tested it and it seems to work robustly, so that's great

PR #473, opened 2026-02-27
handle chat app timeout via config

lucalavezzo:


pmlugato (approved):
great, approved, thanks for the patch @lucalavezzo

PR #480, opened 2026-02-27
A/B Testing

JasonMoho:
# Champion/Challenger A/B Testing Pool

Adds deployment-scoped champion/challenger A/B testing so admins can manage an experiment pool, compare a baseline agent against configured variants, and collect user preference data.

## What It Does

When A/B testing is enabled and the pool is valid, eligible turns can trigger two parallel agent runs:

- one arm is always the configured champion
- the other arm is a randomly sampled challenger from the variant pool
- arm positions (`Response A` vs `Response B`) are randomized per comparison
- users vote `A`, `B`, or `Tie`
- per-variant win/loss/tie metrics are tracked in Postgres

The current implementation is deployment-wide and server-authoritative:

- experiment configuration is stored under `services.chat_app.ab_testing`
- runtime state is persisted in Postgres-backed static config, not in the original YAML after deploy
- sampling decisions are made on the server
- unresolved comparisons are tracked per conversation, with `max_pending_per_conversation` enforced by the backend and UI

## Current Configuration Model

Configure A/B testing under:

```yaml
services:
  chat_app:
    ab_testing:
      enabled: true
      ab_agents_dir: /root/archi/ab_agents
      sample_rate: 0.25
      disclosure_mode: post_vote_reveal
      default_trace_mode: minimal
      max_pending_per_conversation: 1
      target_roles: []
      target_permissions: []
      pool:
        champion: baseline
        variants:
          - label: baseline
            agent_spec: baseline-ab.md
          - label: poet
            agent_spec: poet-ab.md
            provider: openrouter
            model: anthropic/claude-3.5-sonnet
            recursion_limit: 30
          - label: concise
            agent_spec: concise-ab.md
            num_documents_to_retrieve: 3
```

Important differences from the original version:

- A/B config lives only at `services.chat_app.ab_testing`
- A/B variants use `label` plus `agent_spec`, not `name`
- `agent_spec` must be a filename under `ab_agents_dir`
- A/B agent specs are isolated from normal `agents_dir`
- incomplete A/B config no longer blocks startup; it starts inactive and surfaces warnings in the admin UI/runtime

### Participant and Runtime Endpoints

| Method | Endpoint | Description |
|---|---|---|
| `GET` | `/api/ab/pool` | Returns participant-safe pool info, or full admin payload for admins |
| `GET` | `/api/ab/decision` | Server-side decision for whether the next turn should use A/B |
| `POST` | `/api/ab/compare` | Streams a champion-vs-challenger comparison as NDJSON |
| `POST` | `/api/ab/preference` | Submits the user vote (`a`, `b`, `tie`) |
| `GET` | `/api/ab/pending` | Returns unresolved comparisons for a conversation |
| `GET` | `/api/ab/metrics` | Returns aggregate per-variant metrics |

### Admin Configuration Endpoints

| Method | Endpoint | Description |
|---|---|---|
| `POST` | `/api/ab/pool/set` | Legacy full-pool save path |
| `POST` | `/api/ab/pool/settings/set` | Saves experiment settings only |
| `POST` | `/api/ab/pool/variants/set` | Saves variants only |
| `POST` | `/api/ab/pool/disable` | Disables A/B testing while retaining saved variant config |

### Admin Management

The original pool editor in the chat settings modal has been replaced by a dedicated admin page:

- data page top-right action button: `A/B Testing`
- admin route: `/admin/ab-testing`
- page is admin-only

The admin page includes:

- experiment status badge
- sample rate
- disclosure mode
- default agent-activity visibility
- max pending per conversation
- champion selector
- full variant list with:
  - label
  - A/B-only agent spec
  - provider override
  - model override
  - recursion limit override
  - document retrieval override
- split save actions:
  - `Save Configuration`
  - `Save Variants`
- `Disable`
- add/remove variant actions
- inline creation of new A/B-only agent specs from the agent-spec selector

### Participant Experience

For sampled turns:

- both responses stream side-by-side
- the comparison uses neutral `Response A` / `Response B` labels
- variant identity is shown according to `disclosure_mode`
- agent activity visibility is controlled by `default_trace_mode`
- voting UI appears after the comparison is ready

### Pending Comparison Behavior

The implementation now supports queued unresolved comparisons:

- `max_pending_per_conversation` controls how many unresolved comparisons a conversation may hold
- the UI restores all pending comparisons on reload
- input is blocked only when the configured pending limit is reached

## Agent Spec Isolation

A/B testing no longer reuses the normal user agent pool.

- normal user-visible agent specs live in `agents_dir`
- A/B-only specs live in `ab_agents_dir`
- A/B-created specs are hidden from normal user agent CRUD
- when `ab_agents_dir` is configured in source YAML, deployment staging copies those files into the deployment and rewrites runtime config to `/root/archi/ab_agents`

## Admin Access and Auth

The current product path is:

- auth required
- admin page and admin A/B endpoints are RBAC-gated
- temporary basic-auth admin testing overrides were removed

That means:

- basic auth remains identity-only
- real A/B admin access should come from the permanent auth/RBAC path

## Database

Current database support includes:

- `ab_comparisons`
  - stores each comparison
  - variant metadata
  - prompt and response references
  - recorded preference
- `ab_variant_metrics`
  - aggregate wins
  - losses
  - ties
  - totals

## Backend Changes in the Current Version

- `src/interfaces/chat_app/app.py`
  - dedicated A/B admin page route
  - admin and participant A/B APIs
  - server-side A/B decision logic
  - split settings/variants save paths
  - queued pending-comparison handling
  - runtime config refresh from persisted static config
- `src/utils/ab_testing.py`
  - `ABVariant`
  - `ABPool`
  - validation and pool loading
  - isolated `ab_agents_dir` resolution
  - warning-based inactive startup behavior
- `src/utils/conversation_service.py`
  - A/B comparison creation
  - preference submission
  - pending comparison list/count support
  - metrics updates
- `src/utils/sql.py`
  - SQL for A/B comparison and metrics flows
- `src/cli/templates/base-config.yaml`
  - current `services.chat_app.ab_testing` rendering
  - `ab_agents_dir`
  - split current pool fields

## Notable Behavioral Differences from the Original PR Draft

- Admin management is no longer inside the chat settings modal; it is on a dedicated `/admin/ab-testing` page.
- The system no longer uses shared `agents_dir` for A/B variants.
- The admin page saves deployment-wide config through Postgres-backed static config, not in-memory only.
- Sampling is server-side, not frontend-only.
- `max_pending_per_conversation` now supports multiple unresolved comparisons instead of a single hardcoded pending state.
- Incomplete A/B config warns and stays inactive instead of failing deployment startup.
- The temporary basic-auth admin bridge used during development/testing has been removed.

## Test Coverage Status

Automated coverage has been updated to reflect the current implementation rather than the original PR draft. The current test surface includes:

- updated Playwright fixtures for current admin payloads and isolated A/B agent pools
- updated A/B workflow coverage for the dedicated admin page and current selectors
- coverage for queued pending comparisons
- backend unit coverage for admin payloads, disable behavior, and pending-comparison enforcement
- regression coverage proving the temporary basic-auth admin bridge was removed without changing permanent auth behavior

## Notable Follow-Up Fixes

After the original feature landed, a number of follow-up fixes and product-shape changes were made:

- moved A/B management from the chat settings modal to a dedicated admin page
- split admin saves into separate settings and variants save flows
- fixed runtime refresh so admin saves rehydrate from persisted Postgres config instead of stale cached state
- made `archi restart -c ...` reseed deployment config so static-config changes are actually applied on restart
- updated disable behavior so the inactive state persists and re-renders correctly
- changed A/B variant config from `name`-based matching to explicit `label` plus `agent_spec`
- isolated A/B-only agent specs into `ab_agents_dir` instead of sharing normal `agents_dir`
- added staging/rendering support for `ab_agents_dir` in deployment generation
- enabled inline creation of new A/B-only agent specs from the admin variant selector
- moved server-side sampling decisions out of frontend-only `Math.random()` logic
- expanded pending-comparison handling from a single unresolved comparison to queue-aware enforcement using `max_pending_per_conversation`
- fixed A/B compare/runtime bugs around A/B agent directory resolution and variant initialization
- cleaned up A/B arm headers and trace presentation to better match the current chat theme
- fixed agent-dropdown delete confirmation behavior so the menu stays open while confirming
- removed the temporary basic-auth admin-testing bridge once dedicated admin-page work was complete

### DIFF
--- a/src/interfaces/chat_app/app.py
+++ b/src/interfaces/chat_app/app.py
@@ -1647,8 +1647,18 @@ class ChatWrapper:
         # PipelineEventFormatter yielding *accumulated* content (not deltas);
         # the last write per arm is therefore the complete response text.
         arm_results = {
-            "a": {"final_text": "", "error": None, "final_emitted": False},
-            "b": {"final_text": "", "error": None, "final_emitted": False},
+            "a": {
+                "final_text": "",
+                "error": None,
+                "final_emitted": False,
+                "duration_ms": None,
+            },
+            "b": {
+                "final_text": "",
+                "error": None,
+                "final_emitted": False,
+                "duration_ms": None,
+            },
         }
         arm_model_used = {
             "a": f"{arm_a_variant.provider or ''}/{arm_a_variant.model or ''}".strip("/"),
@@ -1695,6 +1705,8 @@ class ChatWrapper:
                         final_text = getattr(output, "answer", "") or formatter.last_text or arm_results[arm_label]["final_text"]
                         arm_results[arm_label]["final_text"] = final_text
                         arm_results[arm_label]["final_emitted"] = True
+                        duration_ms = int((_time.monotonic() - t0) * 1000)
+                        arm_results[arm_label]["duration_ms"] = duration_ms
                         event_queue.put({
                             "type": "final",
                             "arm": arm_label,
@@ -1702,7 +1714,7 @@ class ChatWrapper:
                             "usage": output_meta.get("usage"),
                             "model": output_meta.get("model"),
                             "model_used": arm_model_used[arm_label],
-                            "duration_ms": int((_time.monotonic() - t0) * 1000),
+                            "duration_ms": duration_ms,
                         })
             except Exception as exc:
                 arm_results[arm_label]["error"] = str(exc)
@@ -1745,6 +1757,8 @@ class ChatWrapper:
         arm_b_error = arm_results["b"]["error"]
         arm_a_final_text = arm_results["a"]["final_text"]
         arm_b_final_text = arm_results["b"]["final_text"]
+        arm_a_duration_ms = arm_results["a"]["duration_ms"]
+        arm_b_duration_ms = arm_results["b"]["duration_ms"]
 
         if arm_a_error and arm_b_error:
             yield {"type": "error", "message": "Both A/B arms failed",
@@ -1790,6 +1804,10 @@ class ChatWrapper:
             pipeline_used=self.current_pipeline_used,
         )
 
+        # Persist per-arm latency for analysis by reusing the timing table keyed by message_id.
+        self._persist_ab_arm_timing(arm_a_mid, arm_a_duration_ms)
+        self._persist_ab_arm_timing(arm_b_mid, arm_b_duration_ms)
+
         # Get user prompt message ID if not already stored above
         if not user_prompt_mid:
             user_prompt_mid = self._get_last_user_message_id(context.conversation_id)
@@ -1829,6 +1847,8 @@ class ChatWrapper:
             "is_champion_first": is_champion_first,
             "arm_a_message_id": arm_a_mid,
             "arm_b_message_id": arm_b_mid,
+            "arm_a_duration_ms": arm_a_duration_ms,
+            "arm_b_duration_ms": arm_b_duration_ms,
             "variant_label_mode": self.ab_pool.variant_label_mode,
         }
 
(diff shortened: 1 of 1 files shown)


docs/docs/configuration.md
```diff
@@ -257,6 +287,93 @@ data_manager:
...
+
+### Pool Fields
+
+| Field | Type | Default | Description |
+|-------|------|---------|-------------|
+| `enabled` | boolean | `false` | Enable the experiment pool |
+| `ab_agents_dir` | string | `/root/archi/ab_agents` | Optional legacy import directory for migrating A/B markdown specs into the DB catalog |
+| `sample_rate` | float | `1.0` | Fraction of eligible turns that should run A/B |
+| `disclosure_mode` | string | `post_vote_reveal` | One of `blind`, `post_vote_reveal`, `named` |
+| `default_trace_mode` | string | `minimal` | One of `minimal`, `normal`, `verbose` |
```
haozturk: As far as I understand, the following is the mapping between the config and what's shown in the UI:
- `minimal`: `Hidden`
- `normal`: `Collapsed`
- `verbose`: `Expanded`

What's in the config doesn't make sense to me. I would use in the config what's shown in the UI.

pmlugato:
@haozturk thanks again for the comments, I agree with them. Any UI changes to A/B testing configuration now persist across restarts. I also followed your suggestion and added a `force_yaml_override` option should it be needed. Timer is fixed, and naming is clearer and consistent between config/UI/docs. Thanks for the good testing, let me know if you find anything else or if we're good to go.

PR #494, opened 2026-03-05
Fixed: Align chat history with user identity and fix model state synchronization

### LINKED ISSUE #486: Provider update in the UI isn't propagated to the `model_used` in the DB

### LINKED ISSUE #492: Chat history is tied to the client_id instead of the user_id
If you log in from a different browser, you don't see your chats, because it's a different session.

### LINKED ISSUE #496: Archi shows the currently selected model for all messages even if a different model was used for an old message

Viphava280444:
Hi guys

This PR addresses the following issues,

fixed #492:  
Chat history now follows user_id instead of client_id when SSO is enabled and the user is authenticated. If authentication is not enabled or the user is not logged in, it falls back to  using client_id.

fixed #486
Model override in the UI now correctly records the actual model used in the database instead of always storing the default model. 

fixed #496
Conversation history now displays the per-message model label from the database instead of showing the currently selected model for all messages.       

Thank you

### DIFF
--- a/docs/docs/configuration.md
+++ b/docs/docs/configuration.md
@@ -59,6 +59,43 @@ The main chat interface.
 | `host` | string | `0.0.0.0` | Network binding |
 | `num_responses_until_feedback` | int | `3` | Responses before prompting for feedback |
 | `auth.enabled` | bool | `false` | Enable authentication |
+| `alerts.managers` | list | `[]` | Usernames allowed to create and delete alerts |
+
+#### `services.chat_app.alerts`
+
+Controls access to the [Service Status Board & Alert Banners](services.md#service-status-board--alert-banners).
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| `alerts.managers` | list of strings | `[]` | Usernames granted alert manager access |
+
+Access rules (evaluated in order):
+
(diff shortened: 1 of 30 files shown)


Viphava280444:
> > Thanks Ohm, looks mostly ok to me but there is one issue: If the same user types in the same conversation from two different sessions, e.g. two different browsers, you get this issue [1]. Can you please address this?
> > [1]
> 
> Thank you Hasan for reviewing, I will have a look today.

Hello @haozturk,  The issue has been fixed. I missed one method that needed to pass user_id.
Could you please review it again?

Viphava280444:
> I actually misunderstood your question. You're right that if two users use the same browser they will share the same client id, thus have the same chat history. This is not a likely scenario but it can happen. But in any case, this is not introduced by this PR and this issue existed in Archi for a long time and I'd suggest addressing this when we implement a proper user handling in Archi.
> 
> Currently, we need to keep client_id based conversation fetching for deployments that don't use SSO.
> 
> Can you please make an issue for this so that it doesn't get forgotten?

OK, Yes I will take care of creating the issue.

haozturk:
Can you please resolve the conflict @Viphava280444 ?

PR #504, opened 2026-03-05
fix: restore api keys tab and apply session key to pipeline default m…

### LINKED ISSUE #475: fix api keys tab in chat settings
replacing api keys via settings in chat UI has some bugs, for starters replacing OpenAI key passed at deployment from user side doesn't save and always reverts to admin key. needs more testing.

for now, the tab in settings has been hidden.

harz05:
Fixes #475

## Problems
1. The API Keys tab in chat Settings was hidden (`<!-- commented out -->`) due to a known bug where the session key was ignored.

2. When a user saved a session API key but did NOT explicitly select a provider in the UI (i.e. used the pipeline default model), the session key was never looked up and the deployment admin key was always used instead, reverting any per-user override.

## Fix

**`index.html`** — Uncommented the API Keys nav button and section. Updated the description to clarify keys live in the browser session only.

**`app.py`** — In `get_chat_response_stream()`, extended the session key lookup to also cover the no-provider-selected case: when no explicit provider comes in from the frontend, the code now reads the pipeline's active model from `dynamic_config` (or falls back to `services.chat_app.default_provider`), checks if the user has a session key for that provider, and injects it — so the BYOK key is used transparently even with the default pipeline model.

## Tested
- Set a fake session key (`sk-fake999`) → sent a message → got an auth error from OpenAI confirming the session key was injected
- Cleared session key → chat worked normally with env key
- API Keys tab visible and functional in Settings modal

Screenshot from 2026-03-06 02-38-35

### DIFF
--- a/src/interfaces/chat_app/app.py
+++ b/src/interfaces/chat_app/app.py
@@ -3548,8 +3548,28 @@ class FlaskAppWrapper(object):
 
         # Get API key from session if available
         session_api_key = None
-        if provider and 'provider_api_keys' in session:
-            session_api_key = session.get('provider_api_keys', {}).get(provider.lower())
+        session_provider_keys = session.get('provider_api_keys', {})
+        if provider and session_provider_keys:
+            session_api_key = session_provider_keys.get(provider.lower())
+        elif not provider and session_provider_keys:
+            # No explicit provider selected — check if the pipeline's active provider
+            # has a user-stored session key and inject it transparently so BYOK keys
+            # work even when the default pipeline model is being used (fixes #475).
+            try:
+                dyn = get_dynamic_config()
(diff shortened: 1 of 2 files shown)


src/interfaces/chat_app/app.py
juanpablosalas, line 1: For some reason the clear, button is not working. Even after I clear the fake API key and I see that is taking the `ENV` value, the chat logs fail with the previously entered key:

```
openai.AuthenticationError: Error code: 401 - {'error': {'message': 'Incorrect API key provided: sk-fake999.. You can find your API key at https://platform.openai.com/account/api-keys.', 'type': 'invalid_request_error', 'code': 'invalid_api_key', 'param': None}, 'status': 401}```

juanpablosalas (changes requested):

PR #522, opened 2026-03-15
Fix api access to archi app

hassan11196:
## Summary
- Add Bearer token authentication support for programmatic API access via `Authorization: Bearer <token>` header
- Auto-upsert user into `users` table on Bearer auth to prevent FK violations on `conversation_metadata`
- Fix `AttributeError` on `current_model_used` by referencing `self.chat.current_model_used` instead of `self.current_model_used`
- Return 401 JSON for `/api/` routes instead of redirecting to `/login`


hassan11196:
@pmlugato can you review?

pmlugato (approved):
Looks good to me, thanks for this @hassan11196 , merging now

Issue #533, opened 2026-03-24
service redmine broken


mariadalfonso:
I have 
https://github.com/archi-physics/archi/releases/tag/v2.3.0
+ https://github.com/archi-physics/archi/pull/532
+ https://github.com/archi-physics/archi/pull/527

but with this error 
```
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
/usr/local/lib/python3.10/site-packages/requests/__init__.py:109: RequestsDependencyWarning: urllib3 (2.3.0) or chardet (6.0.0.post1)/charset_normalizer (3.4.2) doesn't match a supported version!
  warnings.warn(
/usr/local/lib/python3.10/site-packages/pydantic/_internal/_generate_schema.py:628: UserWarning: <built-in function any> is not a Python type (it may be an instance of an object), Pydantic will allow any object with no validation since we cannot even enforce that the input is an instance of the given type. To get rid of this error wrap the type with `pydantic.SkipValidation`.
  warn(
Starting Redmine Service
(2026-03-23 15:22:52,279) [src.utils.connection_pool] INFO: Connection pool initialized: min=5, max=20
Traceback (most recent call last):
  File "/root/archi/src/bin/service_redmine.py", line 27, in <module>
    redmine_config = get_services_config().get("redmine_mailbox", {})
  File "/usr/local/lib/python3.10/site-packages/src/utils/config_access.py", line 38, in get_services_config
    return get_static_config().services_config or {}
  File "/usr/local/lib/python3.10/site-packages/src/utils/config_access.py", line 25, in get_static_config
    raise ConfigNotReadyError("Static config not initialized in Postgres.")
src.utils.config_access.ConfigNotReadyError: Static config not initialized in Postgres.
```

mariadalfonso:
Submit instance migrated to new agentic version.
for reference the following configuration is running on may4 2026
tag v2.3.0
+ PR #527, #532, #536, #544, #559 [these "bug fixes" goes in the v2.4.0]
+ PR #549 new feature disable the uploader

PR #538, opened 2026-03-26
LLM Model for redmine

mariadalfonso:
redmine  reads default_provider and default_model from redmine_mailbox config, which are None since are were not sure.
Now 
If redmine_mailbox has its own model config → uses that
If not → automatically picks from chat_app 
No YAML duplication needed

this solve issue https://github.com/archi-physics/archi/issues/533#issuecomment-4120152478

### DIFF
--- a/src/bin/service_mailbox.py
+++ b/src/bin/service_mailbox.py
@@ -27,9 +27,9 @@ factory = PostgresServiceFactory.from_env(password_override=read_secret("PG_PASS
 PostgresServiceFactory.set_instance(factory)
 
 mailbox_config = get_services_config().get("redmine_mailbox", {})
-redmine = redmine.Redmine('Redmine_Helpdesk_Mail') # this name tells redmine class to not initialize archi() class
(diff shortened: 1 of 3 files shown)


lucalavezzo (changes requested):

mariadalfonso:
I will close this PR
The configuration file is very long https://github.com/mit-submit/submit-admin/blob/main/archi/config_submit.yaml essentially has repeated blocks between the chat_app and redmine services, maybe can be optimized, we can think something later    

    default_provider: local
    default_model: qwen3:32b
    tools:
      search_local_files:
        enabled: true
      search_metadata_index:
        enabled: true
      search_vectorstore_hybrid:
        enabled: true
      fetch_catalog_document:
        enabled: true
    providers:
      local:
        enabled: true
        base_url: hostname # make sure this matches your ollama server URL!
        mode: ollama
        default_model: "qwen3:32b" # make sure this matches a model you have downloaded locally with ollama
        models:
          - "gpt-oss:120b"
          - "qwen3:32b"

PR #550, opened 2026-04-13
feat(data-manager): add Indico scraper integration

livaage:
Adds support for scraping Indico events and meeting materials, alongside the existing scrapers.

Scraper (indico_scraper.py)
- Fetches event metadata, contributions, and slide attachments via the Indico REST API.
- Converts PDF/PPTX/PPT/ODP slides to Markdown via MarkItDown; strips embedded <latexit> blocks that inflate chunk counts on formula-heavy slides (slide_converter.py).
- Deduplicates attachments when the same slides are uploaded in multiple formats (e.g. PDF + PPTX): keeps the higher-priority format.
- Detects SSO-protected events and authenticates via CERNSSOScraper.

ScraperManager integration
- collect_indico() / schedule_collect_indico() hooks.
- URL routing: explicit "indico-" prefix in weblists, plus auto-detection for URLs with "indico" in the hostname and /event/ in the path.
- Indico documents use source_type="web" (matching the existing CHECK constraint) with a "scraper": "indico" metadata field for filtering.
- Most of the changes to the scraper_manager.py is just linting 

Vectorstore
- Prepends a one-line metadata header to each Indico chunk (event title, date, contribution, speaker, affiliation, start time, duration, session) so BM25 retrieval can match on speaker name, time of day, etc. Gated on metadata "scraper"="indico"; no other sources affected.

Config / docs / examples
- base-config.yaml: indico source block (disabled by default).
- docs/docs/data_sources.md: Indico section.
- examples/agents/indico-assistant.md: agent spec for Indico queries.
- examples/deployments/basic-agent/indico_example.list: example weblist.
- SourceRegistry: register "indico" source (depends on links).

Dependencies
- pyproject.toml + requirements-base.txt: add markitdown[pdf,pptx].

Known limitations
- Images and figures in slides are not extracted or described; only text content is converted to Markdown. Slides that communicate primarily through plots/diagrams will produce thin or empty chunks.
- LaTeX in slides: embedded <latexit> blocks are stripped (they are base64-encoded and useless for retrieval), but inline LaTeX notation and formula-heavy slides may still produce low-quality Markdown that chunks poorly.
- Slide context is per-page, not per-deck: each chunk comes from one page/section of the converted Markdown. There is no cross-slide summarisation, so a narrative that spans multiple slides may be split across chunks without connecting context.
- Category URLs (/category/<id>/) are handled in the code but not yet tested end-to-end; only event URLs are documented.
- SSO authentication is CERN-specific (CERNSSOScraper). Other Indico instances with different login flows would need a different auth path.
- No rate limiting or incremental scraping; large events with many attachments are processed in a single run.
- indico_scraper.py is large (~1300 lines); a follow-up refactor could
  split API interaction from resource construction.

### DIFF
--- a/docs/docs/data_sources.md
+++ b/docs/docs/data_sources.md
@@ -113,6 +113,40 @@ Once enabled in config, deploy normally with `archi create --config <config.yaml
 
 ---
 
+## Indico
+
+Ingest events, contributions (talks) and slide materials from an [Indico](https://getindico.io/) instance. Slides in PDF/PPTX/PPT/ODP format are converted to Markdown via [MarkItDown](https://github.com/microsoft/markitdown).
+
+### Configuration
+
+```yaml
+data_manager:
+  sources:
+    indico:
+      enabled: true
+      base_url: https://indico.cern.ch
+      use_sso: false       # set true for SSO-protected events
+      slide_conversion:
+        enabled: true
+        formats: [pdf, pptx, ppt, odp]
+```
+
+Add event URLs to your link lists. URLs with `indico` in the hostname and `/event/` in the path are auto-detected. For Indico instances without `indico` in the hostname, use the explicit `indico-` prefix:
+
+```
+https://indico.cern.ch/event/1408515/
+https://indico.stfc.ac.uk/event/1825/
+indico-https://events.example.org/event/42/
+```
+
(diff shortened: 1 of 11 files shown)


pmlugato (approved):
thanks @livaage , great to see this, will get a lot of use from people I think!

Krittin gave this a review and is happy with it, plus will do the minor changes he needs to integrate it into the scraping refactoring he's done in https://github.com/archi-physics/archi/pull/547 .

Good to go into dev for me then! Merging now

PR #556, opened 2026-04-22
Add generic sidecar support for MCP servers

haozturk:
### Summary

Adds first-class support for MCP servers that run as their own Docker containers (sidecars), communicating with the chatbot over HTTP. Previously only stdio MCP servers (running in-process) and external HTTP MCP servers were supported. This lets tool-specific MCP servers ship their own dependencies, credentials, and setup logic in their own repo — without any archi-side code.

### Motivation

CMS CompOps needs a Rucio MCP server that requires heavy dependencies (`rucio-clients`, `voms-clients-java`), credential handling (X509 proxies), and a Rucio config file. Shoving all that into the chatbot container would couple archi to one specific tool. A generic sidecar pattern solves it once for every future tool.

### Changes

Four files, all generic — no tool-specific strings:

- **`base-config.yaml`** — render four new optional MCP server fields: `build_context`, `image`, `env_from_secrets`, `host_file_mounts`
- **`base-compose.yaml`** — when an MCP server declares `build_context` or `image`, emit a compose service for it with the declared env vars, secrets, and host file mounts (supports asymmetric `src → dest` mount form)
- **`templates_manager.py`** — pass the `mcp_servers` config through to the compose template (one-line addition)
- **`mcp.py`** — strip archi-only fields (`build_context`, `host_file_mounts`, etc.) before passing config to `MultiServerMCPClient`, which only knows about transport-specific fields

### New deployment config schema

```yaml
mcp_servers:
  rucio:
    transport: streamable_http
    url: http://localhost:8000/mcp
    build_context: /path/to/rucio-mcp
    env:
      RUCIO_HOST: http://cms-rucio-int.cern.ch
      # ... plain config, checked in
    env_from_secrets: []       # Docker secrets
    host_file_mounts:
      - /etc/pki/tls/certs/CERN-bundle.pem
      - src: /tmp/x509up_u1000
        dest: /tmp/x509up
```

### What this enables

Adding a new MCP server becomes a 3-step recipe with zero archi framework changes:

1. Write the MCP server in its own repo with HTTP transport + Dockerfile such as https://github.com/haozturk/rucio-mcp
2. Declare it in the deployment config with `build_context` + any env/mounts it needs
3. Restart

### DIFF
--- a/src/archi/pipelines/agents/tools/mcp.py
+++ b/src/archi/pipelines/agents/tools/mcp.py
@@ -6,8 +6,9 @@ from langchain_mcp_adapters.client import MultiServerMCPClient
 from langchain_mcp_adapters.tools import load_mcp_tools
 from langchain.tools import BaseTool
 
-from src.utils.config_access import get_mcp_servers_config
+from src.utils.config_access import get_mcp_servers_config, get_full_config
 from src.utils.logging import get_logger
+from src.archi.pipelines.agents.utils.skill_utils import load_skill
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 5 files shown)


juanpablosalas:
#557 was merged with these changes, closing

PR #557, opened 2026-04-24
Generic MCP sidecar + domain-knowledge skills + tool-use observability

haozturk:
## Generic MCP sidecar + domain-knowledge skills + tool-use observability

Bundles four related improvements to make the MCP integration both more flexible and more operable. Framework stays tool-agnostic — all enhancements are generic.

### Summary

**1. Generic MCP sidecar support**

Lets any MCP server declare `build_context` or `image` in its config and run as its own Docker container, communicating with the chatbot over HTTP. Previously only external HTTP MCP servers (like deepwiki) and in-process stdio servers were supported. The compose template now emits a sidecar service block for any MCP server that declares a build or image, with optional `env`, `env_from_secrets`, and `host_file_mounts` (literal path or `{src, dest}`) fields. Supports `host_mode` for hosts that need it.

**2. `skill` field for MCP servers**

Each MCP server can now declare `skill: <name>` in the deployment config. Archi loads the corresponding `.md` file from the skills directory and appends it to the agent's **system prompt** (once), not to each tool's description. This avoids multiplying a ~5 KB skill by the number of tools in the catalog (which would add ~40 K prompt tokens for a 33-tool MCP server). Fields `build_context`, `image`, `env_from_secrets`, `host_file_mounts`, `skill`, `path` are stripped before passing configs to `MultiServerMCPClient` so they don't break the client.

**3. MCP tool-input capture for UI visibility**

MCP tool calls were displaying empty `{}` arguments in the chat UI because LLM tool-call chunks arrive without complete args — they only materialize at tool-execution time. The existing `_store_tool_input` + `RunMemory.record_tool_input` mechanism already handles this for static tools (retriever, local_files); this PR wires the same hook into `sync_wrapper` inside `_build_mcp_tools`, so MCP invocations record their real args into RunMemory and the UI can display them.

**4. Cache-hit telemetry**

`_extract_usage_from_message`/`_extract_usage_from_metadata` now also pull cached-token counts from LangChain's standardized `input_token_details.cache_read` and the OpenAI-native `prompt_tokens_details.cached_tokens`. Aggregated across the turn and logged as `Usage: prompt=X (cached=Y, Z% hit) completion=W total=T`. Enables measuring prompt-caching effectiveness as the system prompt / tool catalog evolves. Non-cache-reporting providers see `cached=0` with no errors.

### Changes

| File | Purpose |
|---|---|
| `src/cli/templates/base-config.yaml` | Render `env`, `env_from_secrets`, `host_file_mounts`, `build_context`, `image`, `skill` fields on MCP server configs |
| `src/cli/templates/base-compose.yaml` | Emit a compose service block for any MCP server with `build_context`/`image`; supports per-server env, secrets, host-file mounts, and `host_mode` |
| `src/cli/managers/templates_manager.py` | Pass `mcp_servers` config through to the compose template |
| `src/archi/pipelines/agents/tools/mcp.py` | Load skills, strip archi-only fields before `MultiServerMCPClient`, return skills text separately so it can be added to the system prompt once |
| `src/archi/pipelines/agents/base_react.py` | System-prompt injection of MCP skills; MCP sync_wrapper records tool input into RunMemory; cache-token fields surfaced in usage extraction; per-turn usage logging |

### Backwards compatibility

- Existing MCP server configs (stdio with `command`/`args`, external HTTP with just `url`) continue to work unchanged.
- Deployments with no `skill` field behave exactly as before.
- Non-OpenAI providers report `cached_tokens=0` with no errors.

### DIFF
--- a/src/archi/pipelines/agents/tools/mcp.py
+++ b/src/archi/pipelines/agents/tools/mcp.py
@@ -6,8 +6,9 @@ from langchain_mcp_adapters.client import MultiServerMCPClient
 from langchain_mcp_adapters.tools import load_mcp_tools
 from langchain.tools import BaseTool
 
-from src.utils.config_access import get_mcp_servers_config
+from src.utils.config_access import get_mcp_servers_config, get_full_config
 from src.utils.logging import get_logger
+from src.archi.pipelines.agents.utils.skill_utils import load_skill
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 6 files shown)


src/archi/pipelines/agents/tools/mcp.py
```diff
@@ -1,41 +1,92 @@
...
+        transport = cfg.get("transport")
+        if transport == "stdio":
+            # stdio subprocesses inherit nothing by default (mcp.client.stdio uses
+            # an empty env). Forward the parent process env so stdio MCP servers see
+            # what they need.
+            cfg["env"] = {**os.environ, **(cfg.get("env") or {})}
+        else:
+            # For HTTP-based transports, `env` is for the sidecar container (compose),
+            # not the MCP client connection — drop it here.
+            cfg.pop("env", None)
```
juanpablosalas, line 58: there might be other MCPs in the future that have to be HTTP based transports and need environment variables. For now I think it's okay to leave it this way but as we continue to develop MCPs we may want to leave it as general as possible, maybe delete cfg["env] only if the configuration env is empty.

juanpablosalas (approved):
I tested this using the rucio-mcp and it works as expected. The OpenAI cached tokens compatibility is nice to have and in general the ability to include add env variables to the MCP clients as well as the skills will be useful for further development with MCPs. I left a small comment about further compatibility to other http-based mcps but it's not a stopper for merging, LGTM!

PR #561, opened 2026-05-07
Dropped Jinja's default(..., true) boolean flag so only undefined values fall back to the default

haozturk:
Explicit 0.0 now passes through instead of being treated as falsy.
Also make the default 0.0


hassan11196 (approved):
LGTM

PR #562, opened 2026-05-07
Fix two A/B streaming crashes that hid the vote buttons

haozturk:
`stream_ab_comparison` blew up mid-stream after both arms finished, so the `ab_meta` event never reached the frontend — both responses rendered but no `comparison_id`, no vote buttons, input stuck disabled.

Two bugs:

Missing `config_name` arg to _prepare_chat_context. The other two callers were updated when this arg became required; this one was missed. Added positionally.
Stray `self.current_pipeline_used` — that attribute only exists on `RedmineMailer` / grader app, not `ChatWrapper`. Looks like copy-paste. Replaced with a local derived from the chat config, same way __init__ does it.
After: ab_meta fires with a real comparison_id and vote buttons render.

### DIFF
--- a/src/interfaces/chat_app/app.py
+++ b/src/interfaces/chat_app/app.py
@@ -1625,6 +1625,7 @@ class ChatWrapper:
             client_sent_msg_ts,
             client_timeout,
             timestamps,
+            config_name,
             user_id=user_id,
         )
         if error_code is not None:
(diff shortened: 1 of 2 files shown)


haozturk:
I take the liberty to merge this to upgrade the compops production instance. @lucalavezzo @pmlugato @mariadalfonso please be aware of this and report back if it causes an issue with the submit instance. We can revert it and rework it out.

hassan11196 (approved):
LGTM

Issue #558, opened 2026-04-27
Database export to CSV not working as expected


juanpablosalas:
The CSV downloaded by the "Export to CSV" button in the database viewer shows a blank csv with headers.

juanpablosalas:
Solved by #563

PR #563, opened 2026-05-07
fix export to csv

juanpablosalas:
Fixes issue #558 by mapping the rows by index instead of by column name to generate csv.


pmlugato (approved):
approved thanks JP!

PR #572, opened 2026-05-20
Helm charts deployment

juanpablosalas:
### 📝 Summary
This PR introduces support for Kubernetes Helm-based deployments to archi, complementing the existing Docker Compose/Podman setups. It also adds a mechanism to dynamically discover and inject runtime resources (extra_agents and extra_tools) into the pipeline environments via local filesystem scanning.

### 🛠 Key Changes
**⚙️ Core Logic & Dynamism**
- Dynamic Extension Discovery: Added scanning mechanisms to dynamically import and register runtime extensions from absolute base paths (/root/archi/...):
- Agents: src/archi/pipelines/__init__.py scans extra_agents and populates the _PIPELINE_EXPORTS catalog, throwing errors on registration name collisions.
- Tools: src/archi/pipelines/agents/tools/__init__.py scans extra_tools and injects functions/classes into the package global namespace.

**🖥️ CLI & Infrastructure Deployment**
- New install Subcommand: Introduced a robust install command within src/cli/cli_main.py allowing operators to build configurations and orchestrate real deployments via Kubernetes.
- Deployment Management: Updated DeploymentManager to support Helm operations, housing methods like create_deployment_templates and helm_install to parse and safely trigger localized upgrades or forced reinstalls.
- Secret Compiling: Added create_secret_template inside SecretsManager to catch local .env variables, convert secrets safely to Base64 syntax, and compile them into a native Kubernetes manifest target (secrets.yaml).
- Conditional Templating Overhaul: Adjusted TemplateManager workflows to swap out Docker Compose routines for chart, values, config_seed, and tools compilation stages whenever context.helm is enabled.

**🌐 Docker images**
- New "universal" docker images have been pushed to a temporary public docker hub (juanpablosalasg/archi-chat) to be used as source images for the helm deployment

You can test first the template generation using:

```
archi install --name <archi_name> --config <config_path> --env-file <env_path> --services chatbot--verbosity 4 --templates-dir <templates_dir> --dry-run
```

`templates_dir` being the directory where the helm charts will be created. Once you're confident about the templates you can run the installation command without the `--dry-run` flag or manually run `helm install`.


src/cli/managers/templates_manager.py
```diff
@@ -297,14 +459,49 @@ def _copy_default_prompts(self, context: TemplateContext) -> None:
                         shutil.copyfile(prompt_file, dst_file)
                         logger.debug(f"Copied default prompt: {prompt_type}/{prompt_file.name}")
 
+    def _helm_render_default_prompts(self, context: TemplateContext) -> None:
```
Viphava280444, line 607: I think the helm path only handles default prompts. Do we have a method to handle the prompt that user will set to a custom path? e.g.
services:
prompts:
chat_prompt: /path/to/my/custom.prompt
condense_prompt: /path/to/my/condense.prompt

src/cli/managers/templates_manager.py
```diff
@@ -188,27 +220,114 @@ def _stage_agents(self, context: TemplateContext) -> None:
...
+        file_path = chart_dir / "templates/chatbot-agents-configmap.yaml"
+        file_path.parent.mkdir(parents=True, exist_ok=True)
+        with open(file_path,"w") as f:
+            f.write(helm_config)
+
+    def _helm_render_skills(self, context: TemplateContext, skills_data: dict) -> None:
+        chart_dir = context.base_dir / "chart"  
+        tmpl = self.env.get_template(HELM_CHAT_CONFIGMAP)  
+        helm_config = tmpl.render(skills=skills_data, archi_name=context.plan.name) 
+        file_path = chart_dir / "templates/chatbot-agents-configmap.yaml"
```
Viphava280444: I think it will override the agents ConfigMap. Shouldn't it be chatbot-skills....?

src/cli/managers/templates_manager.py
```diff
@@ -297,14 +459,49 @@ def _copy_default_prompts(self, context: TemplateContext) -> None:
...
+        defaults_prompts_dir = repo_root / "examples" / "defaults" / "prompts"
+
+        dict_prompts = {}
+        for prompt_type in ["condense", "chat", "system"]:
+            src_dir = defaults_prompts_dir / prompt_type
+            
+            if src_dir.exists():
+                for prompt_file in src_dir.glob("*.prompt"):
+                    with open(prompt_file, 'r') as f:
+                        file_name = os.path.basename(prompt_file)
```
Viphava280444, line 623: I think it loses 2 of 3 default prompts here. The dict key is only the file name, so all 3 default.prompt files become one key and overwrite each other.

src/cli/templates/helm/templates/chatbot/pvc.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  labels:
+    app: {{ name }}-chatbot
+  name:  {{ name }}-chatbot-pvc
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 500Mi
```
Viphava280444: I think it should be in values.yaml

src/cli/templates/helm/templates/data-manager/pvc.yaml
```diff
@@ -0,0 +1,14 @@
...
+    app: {{ name }}-data-manager-pvc
+    service: {{ name }}-chat
+    app: {{ name }}-chat
+  name:  {{ name }}-data-manager-pvc
+spec:
+  accessModes:
+    - ReadWriteMany
+  resources:
+    requests:
+      storage: 100Mi
```
Viphava280444: I think it should be in values.yaml

src/cli/templates/helm/templates/postgres/pvc.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  labels:
+    app: {{ name }}-postgres
+  name:  {{ name }}-postgres-pvc
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 500Mi
```
Viphava280444: I think it should be in values.yaml

src/cli/templates/helm/values.yaml
Viphava280444, line 1: I think we should add resource requests and limits to each deployment.

src/cli/templates/helm/templates/chatbot/deployment.yaml
```diff
@@ -0,0 +1,95 @@
...
+            name: {{ name }}-chat-config
+          name: {{ name }}-chat-config
+        - name: {{ name }}-chat-prompts
+          configMap:
+            name: {{ name }}-chat-prompts
+        - configMap:
+            name: {{ name }}-chat-agents
+          name: {{ name }}-chat-agents
+        - configMap:
+            name: {{ name }}-chat-skills
```
Viphava280444, line 122: chat-skills is declared as a volume here but I don't see it in volumeMounts

src/cli/templates/helm/values.yaml
Viphava280444, line 1: I think service.type should also be a value. Easier to switch from ClusterIP to NodePort for debugging or external access.

src/cli/templates/helm/templates/data-manager/pvc.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: PersistentVolumeClaim
+metadata:
+  labels:
+    app: {{ name }}-data-manager-pvc
+    service: {{ name }}-chat
+    app: {{ name }}-chat
+  name:  {{ name }}-data-manager-pvc
+spec:
+  accessModes:
+    - ReadWriteMany
```
Viphava280444: Why does it use ReadWriteMany?. I think sometimes ReadWriteMany won't bind on most default StorageClasses.

src/cli/templates/helm/templates/chatbot/deployment.yaml
```diff
@@ -0,0 +1,95 @@
...
+            initialDelaySeconds: 900
+            periodSeconds: 30
+          readinessProbe:
+            httpGet:
+              path: /api/health 
+              port: {{ "{{ .Values.chat.port }}" }}
+            initialDelaySeconds: 60
+            periodSeconds: 5
+          ports:
+          - containerPort: {{ "{{ .Values.chat.port }}" }}
```
Viphava280444: this uses the host port, but the app listens on the container port. When they differ the probe fails. Shouldn't it be .Values.chat.externalPort?

src/cli/templates/helm/templates/chatbot/deployment.yaml
```diff
@@ -0,0 +1,95 @@
...
+          configMap:
+            name: {{ name }}-chat-prompts
+        - configMap:
+            name: {{ name }}-chat-agents
+          name: {{ name }}-chat-agents
+        - configMap:
+            name: {{ name }}-chat-skills
+          name: {{ name }}-chat-skills
+        - configMap:
+            name: {{ name }}-chat-tools
```
Viphava280444, line 126: I think these ConfigMaps only exist when configured, so the pod gets stuck ContainerCreating without them. Shouldn't these volumes be optional?

src/cli/managers/secrets_manager.py
```diff
@@ -2,13 +2,17 @@
...
+from jinja2 import Environment
+import base64
 
 from src.cli.service_registry import service_registry
 from src.cli.source_registry import source_registry
 from src.utils.logging import get_logger
 
 logger = get_logger(__name__)
 
+HELM_SECRETS = "helm/templates/secrets.yaml"
```
Viphava280444, line 14: Secrets.yaml is missing.

swinney:
### Code review

Found 4 issues:

1. `get_required_volumes(helm=True)` crashes with `TypeError` when `--gpu-ids` is set. In Helm mode the set is populated with `(service, volume)` tuples, but the GPU branch unconditionally adds the bare string `"archi-models"`, so `sorted()` on a mixed set of tuples and strings raises `TypeError`. `VolumeManager.create_volume_templates` also unpacks each item as `volume[0]`/`volume[1]`, which fails on the bare string. (bug — any `archi install` with GPUs hits this)

https://github.com/archi-physics/archi/blob/b62ed09db08e447c827a095a4ed9dfae6fad1c6a/src/cli/utils/service_builder.py#L95-L106

2. data-manager and postgres deployments render `resources` from `.Values.chat.resources` instead of `.Values.data_manager.resources` / `.Values.postgres.resources`. `values.yaml` defines separate per-service `resources` blocks, so these services silently inherit the chatbot's limits and can't be tuned independently. Likely a copy-paste error. (bug)

https://github.com/archi-physics/archi/blob/b62ed09db08e447c827a095a4ed9dfae6fad1c6a/src/cli/templates/helm/templates/data-manager/deployment.yaml#L39-L41

https://github.com/archi-physics/archi/blob/b62ed09db08e447c827a095a4ed9dfae6fad1c6a/src/cli/templates/helm/templates/postgres/deployment.yaml#L40-L42

3. Helm rendering crashes with `TemplateNotFound` for any enabled service without a Helm template. Both `_stage_service_artifacts` and `create_deployment_templates` iterate all enabled services and unconditionally load `helm/templates/<service>/{service,deployment}.yaml`, but templates only exist for `chatbot`, `data-manager`, and `postgres`. Enabling `grafana` or `grader` (both registered, real services) raises `TemplateNotFound`. (bug)

https://github.com/archi-physics/archi/blob/b62ed09db08e447c827a095a4ed9dfae6fad1c6a/src/cli/managers/templates_manager.py#L489-L496

https://github.com/archi-physics/archi/blob/b62ed09db08e447c827a095a4ed9dfae6fad1c6a/src/cli/managers/deployment_manager.py#L228-L232

4. `--dry-run` still writes files to disk, including base64-encoded secrets. Only `helm_install` is guarded by `if not dry`; `create_secret_template`, `create_volume_templates`, `prepare_deployment_files`, and `create_deployment_templates` all run unconditionally before it. The `create` command exits early in dry mode (the invariant added in commit `9256a9af`), but `install` does not. (bug — secrets written to disk on a dry run)

https://github.com/archi-physics/archi/blob/b62ed09db08e447c827a095a4ed9dfae6fad1c6a/src/cli/cli_main.py#L685-L705

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

swinney:
By all means, :-1: the claude reviewer if you feel the need.  :smile: 

If you want me to have claude fix the bugs, gimme a :+1:   I have not had a chance to run it with minikube yet.   I'm happy to do that, just need a bit more time.

swinney:
### Follow-up: deploy-blocking bug found while testing on a live cluster

I deployed this branch end-to-end (minikube + `archi install` -> `helm upgrade --install`). postgres, data-manager, and the config-seed post-install Job all came up cleanly. The chatbot pod, however, gets stuck in `ContainerCreating` indefinitely:

```
Warning  FailedMount  configmap "<name>-chat-skills" not found
Warning  FailedMount  configmap "<name>-chat-tools"  not found
```

Root cause: in the chatbot volumes, `optional: true` is indented at the **Volume** level, but `optional` is a field of the `ConfigMapVolumeSource` (i.e. it must sit under `configMap:`). As written, the API server silently drops it, so each mount defaults to `optional: false`. When a deployment has no skills/tools (so those ConfigMaps are never rendered), the missing ConfigMaps wedge the pod forever instead of being treated as optional.

https://github.com/archi-physics/archi/blob/b62ed09db08e447c827a095a4ed9dfae6fad1c6a/src/cli/templates/helm/templates/chatbot/deployment.yaml#L88-L103

All four configMap volumes (`prompts`, `agents`, `skills`, `tools`) have the same misplacement; `config`/`agents`/`prompts` happen to render in this config so they mount, but `skills`/`tools` do not and block startup.

Fix is to nest `optional` under `configMap:`, e.g.:

```yaml
        - configMap:
            name: {{ name }}-chat-skills
            optional: true
          name: {{ name }}-chat-skills
```

After applying that, the chatbot pod schedules and the app boots normally. Note this compounds with the earlier finding that empty list-derived ConfigMaps (weblists, and here skills/tools) aren't rendered at all -- either always emit an empty ConfigMap or rely on a correctly-placed `optional: true`.

🤖 Generated with [Claude Code](https://claude.ai/code)

juanpablosalas:
Thanks for the comments @Viphava280444 and @swinney ! I have already implemented them. For the `grafana` and `grader` services I added blank templates as I am not that familiar with them, I would like to maybe understand how they are deployed from someone who is using them. I think that while I figure this out, the main services and structure change can be reviewed.

src/cli/templates/helm/templates/chatbot/configmap.yaml
```diff
@@ -0,0 +1,65 @@
...
+metadata:
+  annotations:
+    use-subpath: "true"
+  name: {{ archi_name }}-chat-skills
+{% endif %}
+{% if prompts %}
+apiVersion: v1
+data:
+  {%- for prompt_filename, content in prompts.items() %}
+  {{ prompt_filename }}: |
```
JasonMoho, line 57: This uses prompt paths as ConfigMap keys. The generated chart has keys like `condense/default.prompt`, but Kubernetes ConfigMap keys cannot contain `/`.

src/cli/templates/helm/templates/data-manager/deployment.yaml
```diff
@@ -0,0 +1,83 @@
...
+          volumeMounts:
+            - name: hf-cache
+              mountPath: /.cache
+            - mountPath: /root/.accounts
+              name: root-accounts
+            - mountPath: /root/data
+              name: {{ name }}-data-manager-pvc
+            - mountPath: /root/archi/configs
+              name: {{ name }}-chat-config
+            - mountPath: /root/archi/weblists
```
JasonMoho, line 53: This mount is always required, but the ConfigMap is only rendered when input lists exist. With no input lists, the pod references a missing ConfigMap.

src/cli/utils/service_builder.py
```diff
@@ -92,13 +92,20 @@ def get_service(self, name: str) -> ServiceState:
...
-                volumes.add(state.volume_name)
+                if helm:
+                    volumes.add((service_name,state.volume_name))
+                else:
+                    volumes.add(state.volume_name)
+
         if self.gpu_ids:
-            volumes.add("archi-models")
+            if helm:
+                volumes.add(("chatbot","archi-models"))
```
JasonMoho, line 107: `--gpu-ids` does not work for Kubernetes yet. It renders a duplicate chatbot PVC and does not add GPU resource requests.

src/cli/managers/templates_manager.py
```diff
@@ -149,19 +163,31 @@ def _build_workflow(self, context: TemplateContext) -> List[Callable[[TemplateCo
...
+            stages.remove(self._stage_source_copy)
+
         return stages
 
     # individual stages
     def _stage_prompts(self, context: TemplateContext) -> None:
         # Copy default prompt templates (condense/, chat/, system/ structure)
-        self._copy_default_prompts(context)
+        if context.helm:
+            self._helm_render_default_prompts(context)
```
JasonMoho, line 184: Configured custom prompt paths are not packaged or rewritten, so they can still point at host-local files inside the pod config.

src/cli/managers/templates_manager.py
```diff
@@ -188,27 +220,114 @@ def _stage_agents(self, context: TemplateContext) -> None:
...
-        self._copy_markdown_directory(
-            src_dir,
-            dst_dir,
-            missing_message=f"Agents directory not found: {src_dir}",
-            empty_message=f"No agent markdown files found in {src_dir}",
-            required=True,
-        )
+
+        if context.helm:
+            for agent_file in list(src_dir.glob("*.md"))+list(src_dir.glob("*.py")):
```
JasonMoho, line 229: In Helm mode, a missing `agents_dir` can silently package no agents and still report success.

src/cli/cli_main.py
```diff
@@ -592,6 +592,123 @@ def evaluate(name: str, config_file: str, config_dir: str, env_file: str, force:
...
+        **other_flags,
+    )
+
+    # Host-side seeding removed; container config-seed handles schema + ingestion before services start.
+    
+    deployment_manager = DeploymentManager(helm=True)
+    use_selenium = config_manager.use_selenium
+    deployment_manager.create_deployment_templates(base_dir, services=service_only_resolved, env=env, name=helm_name, use_selenium=use_selenium)
+    
+    if not dry:
```
JasonMoho, line 712: Dry-run skips Helm but still reports a created deployment with running services.

src/cli/templates/helm/templates/grafana/deployment.yaml
```diff
@@ -0,0 +1 @@
+#TODO
```
JasonMoho: Remove or implement

JasonMoho (changes requested):
Looks fine, it just needs tests.

swinney:
Great PR.  I'm going to use this with minikube for my local dev stuff.

src/cli/managers/templates_manager.py
juanpablosalas: AFAIK, the `archi create` command only handles copying the prompts in the default directory (examples/defaults/prompts) so the helm installation follows this behaviour. Do you think we should change that for both the create and install commands?

src/cli/managers/templates_manager.py
juanpablosalas: This is intended, by default it will still package the `base_react` (default) agent unless an `agents_dir` is set

src/cli/cli_main.py
juanpablosalas: Agree, I think it doesn't make sense to log the deployment success in this context since it's not the CLI command in charge of the deployment but the subsequent helm chart installation. I removed the logging of deployment success.

juanpablosalas:
Hi @JasonMoho,

Thanks for the comments :) I implemented the changes and fixed what was needed to be changed for the unit tests to pass. The build base images action I believe is failing since the PR was not created in the main repo. Should we try to move it to the main ?

JasonMoho (approved):
Merge once conflicts are fixed

PR #573, opened 2026-05-28
Feat: Jira service for drafting ticket answers autonomously

AntuBattle:
This Jira service will allow Archi to answer ticket autonomously polling directly from the cloud platform.

### DIFF
--- a/src/bin/service_jira.py
+++ b/src/bin/service_jira.py
@@ -1,0 +1,56 @@
+#!/bin/python
+import time
+
+from src.interfaces.jira import (
+    JiraIssueClient,
+    JiraServiceConfig,
+    JiraTicketResponderService,
+    build_archi_for_jira,
+    resolve_jira_agent_settings,
+)
+from src.utils.config_access import get_services_config
+from src.utils.env import read_secret
+from src.utils.logging import get_logger, setup_logging
+from src.utils.postgres_service_factory import PostgresServiceFactory
+
+setup_logging()
+logger = get_logger(__name__)
+
+
+def resolve_jira_pat() -> str:
+    pat = read_secret("JIRA_TICKET_RESPONDER_PAT")
+    if pat:
+        return pat
+    raise ValueError("Missing Jira auth: set JIRA_TICKET_RESPONDER_PAT.")
+
+
+def main() -> None:
+    logger.info("Starting Jira ticket responder")
+
+    factory = PostgresServiceFactory.from_env(password_override=read_secret("PG_PASSWORD"))
+    PostgresServiceFactory.set_instance(factory)
+
+    services_config = get_services_config()
+    jira_config = JiraServiceConfig.from_config(services_config.get("jira_ticket_responder", {}))
+    pat = resolve_jira_pat()
+
+    agent_settings = resolve_jira_agent_settings(services_config)
+    archi_instance, agent_settings = build_archi_for_jira(services_config, agent_settings)
+
+    issue_client = JiraIssueClient(jira_config.url, pat)
+    service = JiraTicketResponderService(
+        config=jira_config,
+        issue_client=issue_client,
+        archi_instance=archi_instance,
+        postgres_factory=factory,
+        agent_settings=agent_settings,
+    )
+    logger.info("Jira ticket responder lookback window: %s days", jira_config.lookback_days)
+
+    while True:
+        service.poll_once()
+        time.sleep(jira_config.poll_interval_minutes * 60)
+
+
+if __name__ == "__main__":
+    main()
--- a/src/cli/managers/config_manager.py
+++ b/src/cli/managers/config_manager.py
@@ -8,15 +8,17 @@ import yaml
 from src.cli.managers.templates_manager import BASE_CONFIG_TEMPLATE
 from src.cli.source_registry import source_registry
 from src.utils.ab_testing import ABPool, ABPoolError, load_ab_pool_state
+from src.utils.jira import parse_jira_project_keys
 from src.utils.logging import get_logger
 
 logger = get_logger(__name__)
 
-STATIC_FIELDS = ['global', 'services']
+STATIC_FIELDS = ["global", "services"]
+
 
 class ConfigurationManager:
     """Manages archi configuration loading and validation"""
-    
+
     def __init__(self, config_paths_list: List[str], env):
         self.configs = []
         for config_filepath in config_paths_list:
@@ -25,59 +27,69 @@ class ConfigurationManager:
                 config = self._load_config(config_filepath)
                 self._append(config)
             except Exception as e:
-                logger.error(f'Config {config_filepath} could not be loaded due to {str(e)}')
+                logger.error(
+                    f"Config {config_filepath} could not be loaded due to {str(e)}"
+                )
 
-        assert(len(self.configs)>0)        
+        assert len(self.configs) > 0
         self.config = self.configs[0]
 
         self.env = env
-    
+
     def _load_config(self, config_filepath) -> Dict[str, Any]:
         """Load and validate basic structure of config file"""
         if not config_filepath.exists():
             raise FileNotFoundError(f"Configuration file not found: {config_filepath}")
 
-        with open(config_filepath, 'r') as f:
+        with open(config_filepath, "r") as f:
             config = yaml.safe_load(f)
 
         if not config:
             raise ValueError("Configuration file is empty or invalid")
 
         if "archi" in config:
-            raise ValueError("The 'archi' section is no longer supported in config.yaml.")
+            raise ValueError(
+                "The 'archi' section is no longer supported in config.yaml."
+            )
 
         # Track origin for relative-path resolution (e.g., prompts).
         config["_config_path"] = str(config_filepath)
 
         return config
-    
-    def _append(self,config):
+
+    def _append(self, config):
         """Appends configuration to the config list if the static portions are equivalent to the previous one."""
 
-        if len(self.configs)==0:
+        if len(self.configs) == 0:
             self.configs = [config]
         else:
             previous_config = self.configs[-1]
 
-            #This does not assume the static_fields to be required 
+            # This does not assume the static_fields to be required
             for static_field in STATIC_FIELDS:
                 if static_field in previous_config.keys():
                     if not static_field in config.keys():
-                        raise ValueError(f"The field {static_field} must be present in all configurations.")
+                        raise ValueError(
+                            f"The field {static_field} must be present in all configurations."
+                        )
 
                     if previous_config[static_field] != config[static_field]:
-                        raise ValueError(f"The field {static_field} must be consistent across all configurations.")
+                        raise ValueError(
+                            f"The field {static_field} must be consistent across all configurations."
+                        )
 
             self.configs.append(config)
-    
-    def _get_static_required_fields_for_services(self, services: List[str]) -> List[str]:
+
+    def _get_static_required_fields_for_services(
+        self, services: List[str]
+    ) -> List[str]:
         """Get required configuration fields based on enabled services"""
         if not services:
             return []  # No validation needed if no services selected
 
         # Base fields always required
         requirements = [
-            'name', 
+            "name",
         ]
 
         # Services that have additional required fields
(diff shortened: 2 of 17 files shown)


src/cli/templates/base-config.yaml
```diff
@@ -91,6 +91,16 @@ services:
     default_provider: {{ services.redmine_mailbox.default_provider | default('local', true) }}
     default_model: {{ services.redmine_mailbox.default_model | default('qwen3:32b', true) }}
     providers: {{ services.redmine_mailbox.providers | default({"local": {"enabled": true, "base_url": "http://localhost:11434", "mode": "ollama", "default_model": "llama3.2", "models": ["llama3.2"]}}) | tojson }}
+  jira:
```
lucalavezzo: Minor comment: there is already a `jira` data source. It may be slightly confusing to also have a `jira` service.

src/interfaces/jira.py
```diff
@@ -0,0 +1,431 @@
...
+
+def require_provider_secret(default_provider: str, read_secret: Callable[[str], str]) -> None:
+    secret_name = provider_secret_name(default_provider)
+    if secret_name is None:
+        return
+    if not read_secret(secret_name):
+        raise ValueError(f"Missing required provider secret for Jira service: {secret_name}")
+
+
+def provider_secret_name(default_provider: str) -> Optional[str]:
```
lucalavezzo: I would imagine this should go elsewhere, if we don't have something for this already

src/interfaces/jira.py
lucalavezzo, line 1: there are a lot of nice modular functions here: since we already have a jira collector for tickets, can some of these be shared?

src/cli/service_registry.py
```diff
@@ -151,6 +151,20 @@ def _register_default_services(self):
...
+            name='jira',
+            description='Jira integration service for answering tickets',
+            category='integration',
+            depends_on=['postgres'],
+            required_secrets=['JIRA_SERVICE_PAT'],
+            required_config_fields=[
+                'services.jira.url',
+                'services.jira.project',
+                'services.jira.visible_to_role',
+                'services.jira.poll_interval_minutes',
```
lucalavezzo: I don't think this needs to be required, was there a specific motivation?

src/cli/managers/config_manager.py
```diff
@@ -135,6 +135,7 @@ def validate_configs(self, services: List[str], sources: Optional[List[str]] = N
             self._validate_config(required_fields, config)
             self._validate_chat_app_config(config, services)
             self._validate_benchmarking_config(config, services)
+            self._validate_jira_config(config, services)
```
lucalavezzo, line 157: I actually forgot we have all these checks hardcoded in the CLI .... not great ... but okay, out of scope here 😅

AntuBattle: Oh okay, I thought I was meant to add the validation there too since I saw the other ones 😅
 I will remove it

src/cli/templates/base-config.yaml
AntuBattle: Yeah that's true. Maybe `jira_ticket_service` or  `jira_ticket_answering` since it's already in the services section?

src/cli/service_registry.py
AntuBattle: That's true, let's keep it optional and default it to 1 min as we said. It's not strictly necessary for the user to set it.

src/interfaces/jira.py
AntuBattle: You are right, Archi already fails internally if there is a key that is missing. This was just a fail-fast to ensure everything worked and to eventually have a clear error. For now, I would say I remove it since it's out of scope to harden these fails and eventually we can work on that later 😃

src/interfaces/jira.py
AntuBattle: We can, for the moment I kept them scoped here because in theory these two services have different purposes and different "reasons to change" (one ingests tickets in a certain format, the other is concerned with answering them and is not necessarily interested in having them in the same format, or requires additional data), so coupling certain functions might be bad in the long term even if they share some functionality. As I correct comments from your review I will look a bit deeper into what we can share with the existing jira collectors and come back with some suggestions!

src/cli/managers/config_manager.py
lucalavezzo: no it should absolutely be here! I just don't like we have these things hardcoded (rather than say,

```
for service in requested_services:
    service.validate_config()
```

but don't worry about it

src/interfaces/jira.py
lucalavezzo: Sounds good. I trust your judgement, I haven't looked carefully at either of those modules, I only wanted to make you aware we had the older one for the data source!

src/cli/templates/base-config.yaml
lucalavezzo: yea sure, either sounds good to me +1

src/cli/managers/config_manager.py
AntuBattle: Ok! I'll take a note eventually so we can refactor it in another PR later

src/cli/service_registry.py
```diff
@@ -151,6 +151,19 @@ def _register_default_services(self):
                                     'services.redmine_mailbox.project']
         ))
 
+        self.register(ServiceDefinition(
+            name='jira_ticket_responder',
+            description='Jira integration service for responding to tickets',
+            category='integration',
+            depends_on=['postgres'],
+            required_secrets=['JIRA_SERVICE_PAT'],
```
haozturk: We already had `JIRA_PAT` secret. Is there any reason why we invented a new secret?

src/cli/templates/base-config.yaml
```diff
@@ -91,6 +91,16 @@ services:
     default_provider: {{ services.redmine_mailbox.default_provider | default('local', true) }}
     default_model: {{ services.redmine_mailbox.default_model | default('qwen3:32b', true) }}
     providers: {{ services.redmine_mailbox.providers | default({"local": {"enabled": true, "base_url": "http://localhost:11434", "mode": "ollama", "default_model": "llama3.2", "models": ["llama3.2"]}}) | tojson }}
+  jira_ticket_responder:
+    url: {{ services.jira_ticket_responder.url | default('', true) | tojson }}
+    project: {{ services.jira_ticket_responder.project | default('', true) | tojson }}
```
haozturk: Looks like this only allows a single projects. We need to be able to provide one more than project. Rename this as `projects` and make it accept multiple projects.

src/cli/service_registry.py
haozturk: I think it makes sense to have two PATs, one for read-only injection and one for auto-response. `JIRA_SERVICE_PAT` is confusing. Rename it `JIRA_TICKET_RESPONDER_PAT` or `JIRA_RESPONDER_PAT`.

Add a one-line docs note explaining the split: "JIRA_PAT = Jira data source (read-only ingestion); JIRA_RESPONDER_PAT = ticket responder service (browse + add restricted comments). Use distinct accounts for least privilege."

src/bin/service_jira.py
```diff
@@ -0,0 +1,60 @@
...
+)
+from src.utils.config_access import get_services_config
+from src.utils.env import read_secret
+from src.utils.logging import get_logger, setup_logging
+from src.utils.postgres_service_factory import PostgresServiceFactory
+
+setup_logging()
+logger = get_logger(__name__)
+
+SERVICE_TIMEZONE = timezone(timedelta(hours=2))
```
haozturk: Hardcoded timezone will lead to issues later (when DST starts). Either get the timezone from JIRA or adopt a new approach for getting new tickets rather than relying on service startup time.

src/bin/service_jira.py
```diff
@@ -0,0 +1,60 @@
...
+    issue_client = JiraIssueClient(jira_config.url, pat)
+    service = JiraTicketResponderService(
+        config=jira_config,
+        issue_client=issue_client,
+        archi_instance=archi_instance,
+        postgres_factory=factory,
+        agent_settings=agent_settings,
+    )
+    service_started_at = datetime.now(SERVICE_TIMEZONE)
+    logger.info("Jira ticket responder startup cutoff: %s", service_started_at.isoformat())
```
haozturk: The fact that the cutoff is set of the service_started_at means that if the service is down, it'll skip the tickets that were created during the period service is down. It's not very robust and likely we'll miss tickets in the future because of this. This is not very critical. Nice to address, but it's also ok to leave it as it is and create an issue so that it can be addressed later.

src/cli/templates/base-config.yaml
haozturk: Also make sure that this feature can work with multiple projects ofc

haozturk (changes requested):

src/bin/service_jira.py
AntuBattle: I very much agree that this is brittle. The only reason I put it here is that we explicitly did not want Archi to answer older tickets. The cleanest solution to make this idempotent would be having Archi draft an answer to every single ticket that has not been resolved or closed, including the ones that already exist.

Of course this would mean that the moment we deploy it Archi will draft answers to currently open tickets as well. If this is not what we want, I suppose the easiest thing would be using as cutoff time the date that the PR gets approved for example, or whatever static time we decide, that does not change and ensures that every time we deploy the feature the result is always the same. Let me know what you think.

haozturk: I would avoid a static cutoff date. One option is polling tickets matching the following JQL or similar:
```
status NOT in ("Closed", "Done") AND updated >= -7d
```
and answer tickets that has no comment. 7 days can be changed or be made configurable. Simple enough to implement and robust enough to cover tickets that are opened during a downtime. What do you think of this?

AntuBattle: I think this could work, thanks for the suggestion. I'll try implementing it and will push all the modifications soon

src/interfaces/jira.py
```diff
@@ -0,0 +1,551 @@
...
+from src.archi.utils.output_dataclass import PipelineOutput
+from src.utils.conversation_service import Message
+from src.utils.jira import parse_jira_project_keys
+from src.utils.logging import get_logger
+from src.utils.postgres_service_factory import PostgresServiceFactory
+from src.utils.sql import SQL_CREATE_CONVERSATION, SQL_INSERT_CONVO
+
+logger = get_logger(__name__)
+
+ELIGIBLE_STATUSES = ("Open", "In Progress")
```
haozturk: Move this to the config

haozturk (changes requested):
Thanks Antonio. Looks better now. I requested a small change. Once you address it, it's good to go.

src/interfaces/jira.py
AntuBattle: Done

AntuBattle:
Fixes completed. I also reordered imports and formatted code into a cleaner look. @haozturk if it's all good on your side, then I would just squash commits before merging so we keep main's history clean.

haozturk:
LGTM Antonio, thanks. Yes, please squash and then I'll merge

AntuBattle:
Done! @haozturk you can go ahead and merge. Thanks 😄

pyproject.toml
```diff
@@ -26,7 +26,8 @@ dependencies = [
     "pandas==2.3.2",
     "isort==6.0.1",
     "pre-commit>=4",
-    "psycopg2-binary==2.9.10"
+    "psycopg2-binary==2.9.10",
+    "jira==3.8.0",
```
haozturk: Why is this needed? jira should be available in the base image that Dockerfile-jira depends on: https://github.com/archi-physics/archi/blob/7c89d9fc2c8308627ddb6790a27f3543b7e6f9ef/requirements/requirements-base.txt#L64

haozturk (commented):
I have noticed one more issue @AntuBattle Please remove the jira requirement from the pyproject.toml file, since it's redundant there unless there's a reason I'm missing. 

Sorry for not realizing this before. Once this is addressed, I'll merge it.

pyproject.toml
AntuBattle: Thanks for pointing it out, I had not realized it was duplication. I removed it, now it should be correct

haozturk: thanks merging!

haozturk (approved):

PR #576, opened 2026-06-05
LangFuse integration

juanpablosalas:


src/interfaces/chat_app/app.py
```diff
@@ -583,6 +587,147 @@ def update_config(self, config_name=None):
...
+            "trace_name": "chat-turn",
+            "user_id": resolved_user_id,
+            "session_id": session_id,
+            "metadata": trace_metadata or None,
+            "tags": ["chat_app"],
+        })
+
+        candidate_kwargs = [
+            langfuse_kwargs,
+            #callback_kwargs,
```
JasonMoho, line 658: cleanup

src/cli/service_registry.py
```diff
@@ -91,7 +91,7 @@ def _register_default_services(self):
             category='application',
             requires_volume=True,
             depends_on=['postgres'],
-            required_secrets=[],
+            required_secrets=['LANGFUSE_PUBLIC_KEY','LANGFUSE_SECRET_KEY','LANGFUSE_BASE_URL'],
```
JasonMoho, line 94: should these secrets be required? Might be a bit annoying if we require users to set 3 more env variables if they aren't using langfuse.

JasonMoho (commented):
Thanks JP.

Could you add some documentation on how to use langfuse and update the PR description?

PR #581, opened 2026-06-14
feat: playbooks — reusable per-user instruction packs with sharing and import/export

Viphava280444:
## What this adds

Playbooks are reusable instruction packs, one set per user. You save a procedure once — like "how we triage held jobs" or "our end-of-day report format" — and the agent follows it every time instead of you re-typing it.

Two ways they run:

- **Automatic.** The agent keeps every playbook's name + description in its system prompt, so when your request matches one it loads and applies it on its own. e.g. "summarize this ticket: login loops on sso, prod down" → it picks up the `summarize-ticket` playbook by itself.
- **Manual.** Type `/name args` to force a specific one for that turn, through a `/` autocomplete menu.

The description is what the agent matches on, so it's written to say what the playbook does and when to use it.

## How it works

- **Progressive disclosure.** Only names + descriptions stay in context (cheap — tens of tokens each). A full body loads only when a playbook is actually used, auto or via `/name`. So 100 playbooks cost a short menu per turn, not 100 bodies.
- **Ownership.** Each request resolves one owner: the verified SSO session identity (the OIDC subject) when auth is on, or the browser `client_id` in anonymous mode. Reads and writes are owner-scoped in SQL, and other users' ids never leak in anonymous mode. Bodies from other users are treated as untrusted when loaded.
- **Authoring in chat.** Save / update / delete go through tools. The agent saves only when you ask, shows a preview and waits for confirmation before writing (so nothing gets lost), and refuses instead of inventing results when a playbook calls for a tool or data it doesn't have. Deletes need a confirm too.
- **Panel.** A master-detail Playbooks panel does the same create/edit/delete outside chat, plus enabling or disabling public ones.
- **Export / import.** Export is a zip of `<name>/SKILL.md` files (frontmatter + markdown). Import takes that zip, a single `.md`, or an old JSON backup. Imports always land private, and one bad file doesn't break the batch.

## Sharing (opt-in)

A playbook is private by default, or public. Public ones are visible to everyone and read-only for non-owners — but they don't show up in your agent automatically. You opt in: enable the public playbooks you want, and only then do they join your agent's list and become usable (auto or `/name`). There's a per-user cap on how many you can enable, and your own playbook shadows a public one with the same name.

## Storage

Three Postgres tables, all created or upgraded by an idempotent migration on startup: `playbooks` (the packs), `conversation_playbook_turns` (which playbook ran on which message — drives the chip and re-apply on regenerate), and `user_enabled_playbooks` (the per-user opt-in). 

## Testing

214 unit tests: the tools, the system-prompt listing, SKILL.md round-trips, validation, visibility + opt-in, the save-confirm and refuse-don't-fabricate behavior, and SSO vs anonymous ownership. Plus a manual run against a live deployment — 44 scenarios (REST, security, concurrency, a full browser journey, restart persistence): 43 passed, 1 known pre-existing issue, no security findings.

### DIFF
--- a/requirements/requirements-base.txt
+++ b/requirements/requirements-base.txt
@@ -19,6 +19,7 @@ httpx==0.27.2
 humanfriendly==10.0
 croniter==2.0.5
 langgraph==1.0.2
+langgraph-prebuilt==1.0.8
 langchain-mcp-adapters==0.1.11
 langchain==1.0.3
 langchain_ollama==1.0.0
--- a/src/cli/templates/init.sql
+++ b/src/cli/templates/init.sql
@@ -396,7 +396,7 @@ CREATE TABLE IF NOT EXISTS conversations (
     context TEXT NOT NULL DEFAULT '',
     
     ts TIMESTAMPTZ NOT NULL,
-    
+
     conf_id INTEGER REFERENCES configs(config_id)
 );
 
@@ -628,6 +628,76 @@ GRANT SELECT ON
     ab_variant_metrics,
     migration_state
 TO grafana;
+{% endif %}
+
+-- ============================================================================
+-- 12. PLAYBOOKS (user-authored playbook library)
+-- ============================================================================
+
+CREATE TABLE IF NOT EXISTS playbooks (
+    id          SERIAL PRIMARY KEY,
+    name        VARCHAR(100) NOT NULL,        -- kebab-case slug, unique per owner
+    description TEXT NOT NULL,                 -- one-line "when to use this" hint
+    body        TEXT NOT NULL,                 -- the instructions/knowledge injected on load
+    owner_id    VARCHAR(200) NOT NULL,         -- verified identity (auth) or client_id (anonymous)
+    visibility  VARCHAR(10) NOT NULL DEFAULT 'private',  -- 'private' | 'public' (read-only for others)
+    created_at  TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at  TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+CREATE UNIQUE INDEX IF NOT EXISTS idx_playbooks_owner_name ON playbooks(owner_id, name);
+CREATE INDEX IF NOT EXISTS idx_playbooks_owner ON playbooks(owner_id);
+CREATE INDEX IF NOT EXISTS idx_playbooks_public ON playbooks(visibility) WHERE visibility = 'public';
+
+-- ============================================================================
+-- 13. CONVERSATION PLAYBOOK TURNS (per-turn playbook tracking side table)
+-- ============================================================================
+
+CREATE TABLE IF NOT EXISTS conversation_playbook_turns (
+    message_id    INTEGER PRIMARY KEY REFERENCES conversations(message_id) ON DELETE CASCADE,
+    playbook_name VARCHAR(100) NOT NULL,
+    playbook_id   INTEGER,
+    created_at    TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+-- Per-user opt-in to PUBLIC playbooks owned by others. A user's always-in-context
+-- listing AND their invokable set are their own playbooks PLUS the public ones they
+-- enable here — so the shared public library never bloats every user's prompt by default.
+CREATE TABLE IF NOT EXISTS user_enabled_playbooks (
+    user_id     VARCHAR(200) NOT NULL,                                   -- enrolling user (OIDC sub or client_id)
+    playbook_id INTEGER NOT NULL REFERENCES playbooks(id) ON DELETE CASCADE,
+    created_at  TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    PRIMARY KEY (user_id, playbook_id)
+);
+CREATE INDEX IF NOT EXISTS idx_user_enabled_playbooks_user ON user_enabled_playbooks(user_id);
+
+-- ============================================================================
+-- 14. PLAYBOOK INVOCATIONS (unified usage ledger — both sources, with status)
+-- ============================================================================
+-- One honest row per playbook use, for BOTH the explicit /name path and the
+-- model-invoked (auto) Playbook tool. Distinct from conversation_playbook_turns
+-- (which serves only the chip/regenerate for explicit /name turns). Keep this DDL
+-- textually identical to PlaybookService.ensure_schema. NO owner_id (owner ids
+-- double as access credentials); NO FK on playbook_id (a row must survive the
+-- playbook's deletion).
+CREATE TABLE IF NOT EXISTS playbook_invocations (
+    id              SERIAL PRIMARY KEY,
+    conversation_id INTEGER,
+    message_id      INTEGER,
+    playbook_id     INTEGER,
+    playbook_name   VARCHAR(100) NOT NULL,
+    source          TEXT NOT NULL CHECK (source IN ('explicit', 'auto')),
+    status          TEXT NOT NULL DEFAULT 'ok'
+                    CHECK (status IN ('ok', 'not_found', 'unavailable', 'error')),
+    arm             TEXT,
+    ts              TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+CREATE INDEX IF NOT EXISTS idx_playbook_invocations_name_ts ON playbook_invocations(playbook_name, ts);
+
+-- Grafana reads the usage ledger for dashboards. A dedicated grant (not the
+-- section-11 bulk list) because that block runs before this table exists.
+{% if use_grafana %}
+GRANT SELECT ON playbook_invocations TO grafana;
 {% endif %}
 
 -- ============================================================================
--- a/src/interfaces/chat_app/app.py
+++ b/src/interfaces/chat_app/app.py
@@ -45,7 +45,7 @@ from src.archi.utils.output_dataclass import PipelineOutput
 # from src.data_manager.data_manager import DataManager
 from src.data_manager.data_viewer_service import DataViewerService
 from src.data_manager.vectorstore.manager import VectorStoreManager
-from src.utils.env import read_secret
+from src.utils.env import read_secret, read_or_create_persistent_secret
 from src.utils.logging import get_logger
 from src.utils.config_access import get_full_config, get_services_config, get_global_config, get_dynamic_config
 from src.utils.config_service import ConfigService, StaticConfig
@@ -55,12 +55,23 @@ from src.utils.sql import (
     SQL_LIST_CONVERSATIONS, SQL_GET_CONVERSATION_METADATA, SQL_DELETE_CONVERSATION,
     SQL_LIST_CONVERSATIONS_BY_USER, SQL_GET_CONVERSATION_METADATA_BY_USER,
     SQL_DELETE_CONVERSATION_BY_USER, SQL_UPDATE_CONVERSATION_TIMESTAMP_BY_USER,
-    SQL_INSERT_TOOL_CALLS, SQL_QUERY_CONVO_WITH_FEEDBACK, SQL_DELETE_REACTION_FEEDBACK,
+    SQL_INSERT_TOOL_CALLS, SQL_QUERY_CONVO_WITH_FEEDBACK,
+    SQL_QUERY_CONVO_WITH_FEEDBACK_NO_PLAYBOOKS, SQL_DELETE_REACTION_FEEDBACK,
     SQL_GET_REACTION_FEEDBACK,
     SQL_CREATE_AGENT_TRACE, SQL_UPDATE_AGENT_TRACE, SQL_GET_AGENT_TRACE,
     SQL_GET_TRACE_BY_MESSAGE, SQL_GET_ACTIVE_TRACE, SQL_CANCEL_ACTIVE_TRACES,
 )
+from src.utils.playbook_service import (
+    PlaybookService, PlaybookNotFoundError,
+    playbook_invocation_text, resolve_playbook_owner,
+)
+from src.archi.pipelines.agents.tools.playbook_tools import (
+    set_playbook_owner, get_playbook_owner,
+    set_pending_playbook, get_pending_playbook, clear_pending_playbook,
+    classify_playbook_tool_result,
+)
 from src.interfaces.chat_app.document_utils import *
+from src.interfaces.chat_app.playbook_routes import register_playbooks
 from src.interfaces.chat_app.service_alerts import (
     register_service_alerts, get_active_banner_alerts, is_alert_manager,
 )
@@ -77,7 +88,9 @@ from src.utils.ab_testing import (
     normalize_ab_trace_mode,
     resolve_ab_agents_dir,
 )
-from src.interfaces.chat_app.event_formatter import PipelineEventFormatter
+from src.interfaces.chat_app.event_formatter import (
+    PipelineEventFormatter, build_playbook_applied_events, playbook_loads_from_events,
+)
 from src.utils.conversation_service import ConversationService
 from src.utils.user_service import UserService
 from src.utils.ab_agent_spec_service import ABAgentSpecService, ABAgentSpecRecord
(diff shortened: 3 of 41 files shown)


src/utils/playbook_service.py
```diff
@@ -0,0 +1,436 @@
...
+    (session user's email/sub/name) is the owner and any request-supplied client_id
+    is ignored — this closes the IDOR. Otherwise (anonymous / auth-disabled) the
+    request client_id is the owner. Returns (owner, error_message); error_message is
+    a string when the request is rejectable, else None.
+    """
+    if auth_enabled and logged_in:
+        su = session_user or {}
+        # The session stores the OIDC subject under 'id' (see sso_callback); 'sub' is a
+        # harmless extra fallback.
+        verified = su.get("email") or su.get("sub") or su.get("id") or su.get("name")
```
AntuBattle: Is the `name` field guaranteed to be unique?

src/interfaces/chat_app/app.py
```diff
@@ -1148,6 +1167,87 @@ def prepare_context_for_storage(self, source_documents, scores):
 
         return context
 
+    def _ensure_playbook_schema(self) -> None:
```
AntuBattle: Here you say this function is best effort. So, if it fails, the app should still start. However, if you take a look at `sql.py` you also modified this: `SQL_INSERT_CONVO = """
INSERT INTO conversations (
    archi_service, conversation_id, sender, content, link, context, ts,
    model_used, pipeline_used, playbook_name
)`

I am imagining this scenario:
- Old DB has conversations without playbook_name.
- App starts.
- Runtime migration fails for any reason: permission, lock, etc.
- App logs warning but continues.
- First chat insert fails with something like:
`column "playbook_name" of relation "conversations" does not exist`

That can break chat for all users after deployment, I think? 

I would say that if  the playbook table is required for inserting conversations (hence the app should not start without it), then let's just make the app fail fast if migration does not work for any reason. Otherwise we should reconsider this into something opt-in (for example using a different DB table instead of modifying the existing conversations). Better explanation [here](https://github.com/archi-physics/archi/pull/581#discussion_r3420590758)

src/archi/pipelines/agents/base_react.py
```diff
@@ -1011,20 +1035,93 @@ def _init_prompts(self) -> None:
...
         tools: List[Callable] = []
+        resolved: set = set()
         for name in tool_names:
+            if name in _REMOVED_PLAYBOOK_TOOL_NAMES:
+                logger.info(
+                    "Tool '%s' was removed: the playbook listing is always in context and the "
+                    "Playbook tool loads playbooks on demand.", name,
+                )
+                continue
+            if name in _LEGACY_PLAYBOOK_TOOL_ALIASES:
```
AntuBattle: I suppose these are compatibility shims from earlier iterations? Same goes for `_REMOVED_PLAYBOOK_TOOL_NAMES`.
I could not see any trace of older code related to this section in the main branch so that is why I am supposing they come from earlier iterations in this branch. If that is the case, I guess we can remove them to improve readability, unless they have a specific purpose I am missing.

src/utils/playbook_service.py
```diff
@@ -0,0 +1,436 @@
...
+    visibility: str = "private"  # 'private' (owner only) or 'public' (whole deployment, read-only for others)
+    created_at: Optional[str] = None
+    updated_at: Optional[str] = None
+
+
+VISIBILITY_VALUES = ("private", "public")
+
+
+def _normalize_visibility(visibility):
+    """Accept the legacy 'team' value (older exports/clients) as 'public'."""
```
AntuBattle: Where is the legacy 'team' used? is it a shim left from older iterations?

tests/unit/test_playbook_tools.py
```diff
@@ -0,0 +1,543 @@
...
+    agent = BaseReActAgent.__new__(BaseReActAgent)
+    agent._playbook_service = None
+    reg = agent.get_tool_registry()
+    assert {"save_playbook", "update_playbook", "delete_playbook"} <= set(reg)
+    out = reg["save_playbook"]().invoke({"name": "x", "description": "d", "body": "b"})
+    assert "unavailable" in out.lower()
+
+
+def test_legacy_spec_tool_names_alias_to_playbook_tools():
+    # Interim builds shipped skill-named tools; specs written against them must
```
AntuBattle: If the compatibility shims in the code can be removed, then this and other related tests can be removed too.

src/utils/playbook_service.py
```diff
@@ -0,0 +1,436 @@
...
+        raise PlaybookValidationError(f"SKILL.md frontmatter is not valid YAML: {exc}") from exc
+    if not isinstance(front, dict):
+        raise PlaybookValidationError("SKILL.md frontmatter must be a YAML mapping")
+    metadata = front.get("metadata") if isinstance(front.get("metadata"), dict) else {}
+    visibility = metadata.get("visibility") or front.get("visibility")
+    return {
+        "name": str(front.get("name") or fallback_name or "").strip(),
+        "description": str(front.get("description") or "").strip(),
+        "body": "\n".join(lines[idx:]).strip(),
+        "visibility": "public" if visibility in ("public", "team") else "private",
```
AntuBattle: Here as well, is "team" actually different from public? If it collapses to just public then it might just as well not be there. Or has this implementation already been deployed somewhere and we need to keep back compatibility?

src/interfaces/redmine_mailer_integration/redmine.py
```diff
@@ -106,9 +106,13 @@ def insert_conversation(self, issue_id, user_message, archi_message, link, archi
...
-                (service, issue_id, "User", user_message, '', '', ts, self.current_model_used, self.current_pipeline_used),
-                (service, issue_id, "archi", archi_message, link, archi_context, ts, self.current_model_used, self.current_pipeline_used),
+                # Columns must match SQL_INSERT_CONVO:
+                # (archi_service, conversation_id, sender, content, link, context, ts,
+                #  model_used, pipeline_used, playbook_name) -- same ts for both just to
+                # have, not as interested in timing info for redmine service; playbooks
+                # are a chat-app feature, so playbook_name is always None here.
+                (service, issue_id, "User", user_message, '', '', ts, self.current_model_used, self.current_pipeline_used, None),
+                (service, issue_id, "archi", archi_message, link, archi_context, ts, self.current_model_used, self.current_pipeline_used, None),
             ]
```
AntuBattle, line 112: I don't think this is good. The fact that you had to change every single line where conversations are introduced (and you will eventually need to change the JIRA integration too, and so on for every single service that will not use playbooks) tightly couples together features that should not know of each other's existence. Furthermore, it introduces bugs because you have to change every single query that deals with conversations table for compatibility. For example, `src/utils/conversation_service.py`, line 160, it still inserts only 9 fields without the playbook. It doesn't look like it is being used in the flow at the moment, but it's just an example of why tight coupling is dangerous.

My take: Instead of modifying the existing `conversations` table, simply create an additional table, for example 
```
conversation_playbook_turns (
    message_id INTEGER REFERENCES conversations(message_id),
    playbook_name VARCHAR(100),
    playbook_id INTEGER NULL
)
```

and use that in the chat app to insert informations about playbooks. This way, you don't need to change the generic SQL query `SQL_INSERT_CONVO` in `sql.py`, reducing the "blast radius" of modifications you need to make.  You also don't need to change anything else of the existing logic and queries to the original conversations table and pre-existing services, and the feature remains scoped to the chat app.

Alternatively, if you don't want to add a new table, avoid modifying the generic SQL query altogether so that the playbook field already defaults to `NULL`, and you don't have to change every existing insertion logic. In doing so, you would just need to use a different SQL query in the **chat app only** to insert playbook information when available, without changing anything else. (This second fix is simpler, but it's generally more brittle and it's bad practice not to state every single column of an `INSERT` clause. Still, I think it is better than modifying every existing insertion logic.)

requirements/requirements-base.txt
```diff
@@ -19,6 +19,7 @@ httpx==0.27.2
 humanfriendly==10.0
 croniter==2.0.5
 langgraph==1.0.2
+langgraph-prebuilt==1.0.8
```
AntuBattle, line 22: Is this needed for the playbooks? Or was it some other fix?

src/archi/pipelines/agents/base_react.py
```diff
@@ -20,11 +20,30 @@
...
+# Interim builds named these tools save_skill/update_skill/delete_skill; specs
+# written against them keep working.
+_LEGACY_PLAYBOOK_TOOL_ALIASES = {
+    "save_skill": "save_playbook",
+    "update_skill": "update_playbook",
+    "delete_skill": "delete_playbook",
+}
+# Replaced by the always-in-context playbook listing + the auto-registered Playbook tool.
+_REMOVED_PLAYBOOK_TOOL_NAMES = {"list_playbooks", "load_playbook"}
+
```
AntuBattle: If we don't need these for back compatibility with something already deployed or that I am missing, I would say we can safely remove them (i.e. if they are just shims from previous iterations in this very branch).

AntuBattle (commented):
First review.
The implementation looks fine, left comments where I thought we could simplify or improve the implementation, or where I had doubts.

Some general considerations:
- Do we want to couple playbooks to `BaseReActAgent`? This means we cannot opt-out of them for any other service (for example, the redmine service ended up using them **and** instantiating the `PlaybookService` even though it probably does not need them? If they are a feature of the chat app only, then they should be opt-in for `CompOpsAgent` via some sort of extension or inheritance, or at least confined to the `CompOpsAgent` subclass only. Otherwise we can leave as is.

- The implementation currently does not work with A/B testing because of `ContextVars`. They reset by default within the `Thread` execution space used by A/B testing execution flow. So we might want to look into that.
    ...

src/cli/templates/init.sql
```diff
@@ -396,7 +396,11 @@ CREATE TABLE IF NOT EXISTS conversations (
     context TEXT NOT NULL DEFAULT '',
     
     ts TIMESTAMPTZ NOT NULL,
-    
+
+    -- Name of the user-invoked playbook applied to this (user) turn, if any.
+    -- The playbook body is injected only in-flight to the model; `content` stays clean.
+    playbook_name VARCHAR(100),
```
nausikt: Thanks for the impressive features! 

One minor addition to Antonio, the PR was close to being an idempotent startup migration in DB perspectives except this line which may works well on fresh install but it will likely be skipped in case of production/running instances where `conversations` table already exists.


## Sol 0
P.S. Avoid touching the conversations table, as Antonio suggested.

## Sol 1
If we inevitably have to modify any table, one possible solution would be change it to `ALTER TABLE ... ADD COLUMN IF NOT Exists` [1] instead after `CREATE TABLE ... conversations` . this way the `init.sql` will remain idempotent startup migration.
[1]
```sql
ALTER TABLE conversations
ADD COLUMN IF NOT EXISTS playbook_name VARCHAR(100);
```
Otherwise, i suspect that the `playbook_name` column may missing and silently break the feature in running instances as Antonio mentioned.

## Sol 2

Since, we're not using full-brown web framework, thus we may lack handful of built-in practices/conventions for Migration layer either Autogenerate/semi-auto-via-cmd/manually-create migration mechanisms. 
*(this better been done in another issue/PR.)*

Where they typically introduce `schema_migrations`, `migrations/*.{sql|py|php}`, `cmdline` for managing/tracking migrations of any deployed instances. These are used to alter/delete column, tear up/down table in sequence by dev timestamp date aligned with features, so every dev/staging/prod env can synchronize their schema seamlessly when data models (schema) were changed.
> See also  [Django migrations](https://docs.djangoproject.com/en/6.0/topics/migrations/#workflow) and [Laravel migrations](https://laravel.com/docs/13.x/migrations#generating-migrations)

So, Alternatively, we could introduced and run `cli/templates/migrations/2026_06_17_add_playbook_name_to_conversations.sql` with the same sql in [1] after `init.sql`. This would leave our community some minimal clue for further migrations for now.

P.S. `migration_state` table is quite close to what we're looking for, but its seems to take care of only just config data migrations idempotent rather than schema migrations in general.

requirements/requirements-base.txt
Viphava280444: It's not a playbook thing, it's a version skew. I pinned it to the same version we have in our current public base image. The previous commit just showed the bug when CI runs, it reloads the requirements and gets a langgraph-prebuilt higher than 1.0.8, higher than the one in our base image, and found out it breaks. So we need to pin it to this version to avoid breaking the other services.

src/interfaces/chat_app/app.py
```diff
@@ -1,9 +1,12 @@
+import io
```
haozturk: it looks like we're bloating this file too much and we should modularize it. See the details:

### Modularize the playbook code out of `app.py`

This PR adds ~650 lines to `app.py`, which is already large. Most of it isn't
chat-flow logic and can be extracted — and we already have the pattern for it:
`service_alerts.py` registers its routes via a Flask Blueprint
(`register_service_alerts(app, ...)`). Please follow that convention:

1. **New `playbook_routes.py` Blueprint** — move the 9 `/api/playbooks*` handlers
   plus the zip import/export parsing there, wired via `register_playbooks(app, ...)`
   like `service_alerts`. (~400 lines out of `app.py`.)
2. **Move `_ensure_playbook_schema` into the init/migration path** rather than running
   `CREATE TABLE` inside `ChatWrapper.__init__`.
3. **Move `_insert_playbook_turn` / `_last_user_playbook_name` into `PlaybookService`** —
   they're raw SQL that shouldn't live on the Flask wrapper.
4. **Keep** the `_prepare_chat_context` / `insert_conversation` integration in `app.py` —
   that part genuinely belongs to the chat flow.

Not a correctness issue (code works and is tested) — purely maintainability and
consistency with the existing Blueprint pattern.

src/interfaces/chat_app/templates/index.html
```diff
@@ -135,9 +135,10 @@ <h1 class="visually-hidden">archi Chat</h1>
                   </svg>
                   New agent
                 </button>
+                <button class="agent-dropdown-playbooks" type="button" role="menuitem">Playbooks…</button>
```
haozturk: We should move the playbook view to the settings view here. This is the natural place for it instead of the agent view.

Image

src/interfaces/chat_app/static/chat.js
```diff
@@ -4964,6 +5369,130 @@ const Chat = {
...
+    const matchesQuery = (s) => !q
+      || s.name.toLowerCase().includes(q)
+      || (s.description || '').toLowerCase().includes(q);
+    const mine = all.filter(s => s.is_enabled !== false).filter(matchesQuery);
+    const pub = all.filter(s => s.is_enabled === false).filter(matchesQuery);
+
+    const tab = Chat._panelTab === 'public' ? 'public' : 'mine';  // default to My playbooks
+    if (tabsEl) {
+      tabsEl.innerHTML =
+        `<button class="playbooks-tab${tab === 'mine' ? ' active' : ''}" data-tab="mine" type="button" role="tab">My playbooks (${mine.length})</button>`
```
haozturk: I find this view confusing. My playbooks imply that you own them but in reality they are simply the active playbooks. The following is more intuitive:
```
[ Active ]   [ Add from public (0) ]

Active
  Yours
    ⚡ transfer-check-cern   public        [Edit] [Delete]
  Added from public
    ⚡ rucio-triage          public · alice [View] [Remove]
```
Looks like this 

Image

src/interfaces/chat_app/templates/index.html
```diff
@@ -368,6 +370,46 @@ <h3 id="agent-spec-title">New Agent</h3>
...
+          <select id="playbook-visibility" class="playbook-visibility-input" aria-label="Playbook visibility">
+            <option value="private">Private (only me)</option>
+            <option value="public">Public (everyone on this deployment can use it)</option>
+          </select>
+          <textarea id="playbook-body" class="playbook-body-input" rows="10" placeholder="the instructions (markdown; $ARGUMENTS receives /name arguments)"></textarea>
+          <div class="playbooks-status" id="playbooks-status" role="alert"></div>
+        </div>
+
+        <div class="playbooks-footer" data-when="list">
+          <button class="playbooks-new" type="button">New playbook</button>
```
haozturk: these buttons look archaic. Please give them a view that goes well with the style of the app.

haozturk (changes requested):

haozturk (changes requested):
Thanks a lot Ohm. I see that all the issues I previously raised are addressed. I realized a couple of other things to address:
1. When the playbook was automatically picked up, it wasn't shown in the tool call as opposed to explicit triggered. Also, it wasn't clear which playbook was fetched and what the details were. I made a PR for this. Please take a look: https://github.com/Viphava280444/A2rchi/pull/15
2. Ensure playbook calls are persisted in the DB. I don't know if this is already the case or not for both auto pickup as well as explicit playbook call. This way, we can see which playbooks are actually useful.
3. Check if these changes are compatible with A/B or not. I tested quickly and there was a failure.
4. When everything is done, please squash the commits.

Viphava280444:
Thank you so much @haozturk 
1.  Thank you for the PR — it looks good to me, and I've merged it.
2. I have checked and can confirm that all playbook calls are persisted in the DB. (see image below)
3. I have fixed the A/B and tested it — it works well.
4. I have squashed all the commits.

Screenshot 2026-07-13 at 13 56 09

haozturk (approved):
LGTM now, thanks Ohm! The CI is failing b/c the PR is open from a fork and GH doesn't allow the usage of the required secrets to push the built image (As it should do). Pushing the image to a registry is unnecessary IMO. I think we should merge `build-base-images` and `preview` jobs so that we don't have push any image anywhere for testing. This is not a blocker for this PR. I manually tested this locally and it works.

I made an issue for it: https://github.com/archi-physics/archi/issues/592

PR #583, opened 2026-06-22
Parallel tool calls + .mcp.json

NamN8Tr:
## Fixes

**`RecursionError` on every MCP tool call** (3a748da)

langchain-mcp-adapters (0.1.11) sets each tool's `args_schema` to the MCP
server's raw JSON-schema *dict* rather than a pydantic model. langchain-core
(1.2.13) then calls `get_all_basemodel_annotations()` on it, which recurses on
`get_origin(cls)` — for a dict (then `None`) that never terminates.

It's caught and logged at DEBUG so it's non-fatal, but it burns ~1000 stack
frames per call and floods the logs. We install a guarded
`get_all_basemodel_annotations` at MCP-client init that adds the missing base
case: a non-type with no generic origin returns no annotations instead of
recursing. Real pydantic `args_schema` models are unaffected. The patch is
idempotent and safe to call on every init.

**Concurrent tool calls corrupt the shared MCP session** (dc0df0e, 390564a)

All MCP tools share one client/session on a single background loop, so when the
model emits parallel tool calls they hit that session at once — corrupting
response routing and leaking server-side connections.

This needed locks on *both* paths, which is why it's two commits: the threading
lock covers the sync wrapper (`stream`), but `astream` invokes the tool
coroutine directly on the MCP session loop, where parallel calls run as
concurrent coroutines rather than blocking threads and so slip past a threading
lock entirely. The second commit wraps the coroutine in an asyncio lock and
moves tool-input recording inside it, so both paths record exactly once.

## Feature: `.mcp.json` server definitions (674e8ad)

MCP servers can now be declared in a `.mcp.json` next to the deployment config —
the project-scoped format Claude Code uses — instead of or alongside the YAML
`mcp_servers:` block.

- `src/utils/mcp_json.py` owns the format: parses `{"mcpServers": ...}`, maps
  Claude `type` (stdio/http/sse) onto langchain transports (`http` →
  `streamable_http`), infers type from `command`/`url` when omitted, validates
  per-transport required fields, and drops `_`-prefixed comment keys.
- archi-only fields (`path`, `host_file_mounts`, `env_from_secrets`,
  `build_context`, `image`, `skill`) pass through untouched, so a single file
  can serve both archi and Claude Code.
- `ConfigurationManager` discovers `.mcp.json` beside each config file (or the
  file named by an explicit `mcp_servers_file:` key) and merges it into
  `config['mcp_servers']`. **YAML entries win on name collision**, with a
  warning, so a deployment config can override individual shared servers.
- `${VAR}` / `${VAR:-default}` placeholders expand at connect time against the
  container env — so secrets in `url`/`headers`/`env` come from the environment
  instead of being baked into rendered configs. Expansion runs *before* the
  stdio `os.environ` merge, so only declared values are expanded, never the
  inherited host env. An unset variable with no default disables that server
  with a logged error rather than failing the whole client.
- `base-config.yaml` now renders `headers`, so HTTP auth survives into the
  seeded runtime config.

### DIFF
--- a/docs/docs/agents_tools.md
+++ b/docs/docs/agents_tools.md
@@ -199,39 +199,16 @@ Archi supports the [Model Context Protocol (MCP)](https://modelcontextprotocol.i
 
 ### How It Works
 
-1. MCP servers are defined in your deployment runtime configuration
+1. MCP servers are declared in a `.mcp.json` file next to your deployment config
 2. An agent spec includes `mcp` in its `tools` list to opt in
 3. At agent initialization, Archi connects to the MCP servers and discovers available tools
 4. The MCP tools are added to the agent's toolset alongside built-in tools
 
 ### Configuration
 
-Define MCP servers in your deployment configuration:
-
-```yaml
-mcp_servers:
-  my_server:
-    transport: "stdio"
-    command: "uvx"
-    args:
-      - "mcp-server-example"
-  web_search:
-    transport: "sse"
-    url: "http://localhost:8080/sse"
-```
-
-Each server entry follows the format expected by the `langchain-mcp-adapters` library:
-
-- **`transport`**: Communication method — `"stdio"` (subprocess), `"streamable_http"` (HTTP), or `"sse"` (HTTP Server-Sent Events)
-- **`command`** / **`args`**: For `stdio` transport, the command to launch the server
-- **`url`** / **`headers`**: For HTTP-based transports, the server endpoint and optional request headers
-
-### `.mcp.json` (Claude-compatible)
-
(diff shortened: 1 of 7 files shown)


examples/agents/submit-cluster.md
lucalavezzo, line 1: this should be in a submit specific repo

examples/agents/cms-comp-ops.md
lucalavezzo, line 1: please don't modify this either

src/archi/pipelines/agents/cms_comp_ops_agent.py
lucalavezzo, line 1: please remove the builtin mcp servers

PR #586, opened 2026-06-30
Feat(jira_responder_service): Answer comments when mentioned

AntuBattle:
This feature allows Archi to see mentions in JIRA tickets, and respond accordingly with full context from comments and issue description.

It extends the existing `jira_ticket_responder` service, adding new functionality.

## Extended Folder Structure with `services` folder
Due to the increasing length of the `service_jira.py` file in the `bin` folder, I eventually had to separate helper functionality in different files. The most logical way of placing these files, to me, was creating a `services/jira_ticket_responder` folder within `src`. A service is, to me, a piece of functionality that, in our case, abstracts interaction with external tools or apis (through code in the `interfaces` folder) and uses that data to perform actions and interact with the database through `factory` objects.

Therefore, it made sense to me to extend the previous file structure that had
`interfaces`
`bin`
`utils`

as core folders with an additional `services` folder with the purpose of storing all those functionalities used in the `bin/**_service.py` files without cluttering the `bin` folder.

## Functionality
Archi will be able to answer mentions strictly mentioning the account configured in its `JIRA_TICKET_RESPONDER_PAT` when deployed.
It will have access to context from other comments and issue description/title as well.
There is no dynamic context management as it is unlikely that issues ever exceed the whole context window of the model.
There is however a safety check that compares the total prompt size (which is equal to system prompt + issue + comments) against the known model context window in the same fashion that `base_react.py` currently does for multi-turn chat. This is just a check that eventually excludes older comments if the prompt exceeds such length, but there is no context compression due to it being deemed unnecessary during previous discussions on the matter.

### Database Migration
This feature adds a new database table to store mentions and issues Archi has already responded to, that allows to avoid answer duplication after a service restart, and addresses concurrency on non-distributed deployments.
Due to Archi not having a database migration standard in place yet, `jira_ticket_responder` service simply performs a check at startup to verify if the table exists, eventually creating it if necessary. This is marked by a comment in the code and will be removed once there is a migration standard in place. 

## Config
A simple `respond_to_mentions` opt-out config field has been introduced.


haozturk:
Thanks a lot for this @AntuBattle Before looking into it deeply, I realized that anyone who has access to the configured JIRA projects can trigger a response with a mention. We shouldn't allow everyone to do it, at least not in the beginning. Can we update this feature in such a way only mentions from accounts with certain roles (say Developers) will be replied?

AntuBattle:
Sure, absolutely. I will modify it so that visibility can eventually be made public but mentioning can still be restricred only to certain roles, I will do it with a simple config variable

AntuBattle:
@haozturk Done. Mentions can now be configured to work only for certain roles specified in config. The same thing is true for comment visibility. Both fields are optional, so if not specified it will just default to all. This way there is no need to modify anything once we remove restrictions eventually.

PR #588, opened 2026-07-06
feat: authenticated MCP support — built-in MCP server + per-user SSO tokens for external MCP servers

hassan11196:
# Authenticated MCP support

This PR lets an archi deployment **speak MCP in both directions, with real authentication**:

1. **archi as an MCP server** — MCP clients (Claude Desktop/Code, VS Code, Cursor) connect to a deployment at `/mcp/sse` and use its RAG pipeline, documents, and stats as tools.
2. **archi as an MCP client, per user** — the agent connects to protected external MCP servers using an OAuth token that belongs to the logged-in user, obtained with one click.
3. **archi as an MCP client, headless** — interfaces with no user session (Mattermost bot, benchmarking, cron) use a service-account token instead.

A new **Settings → MCP Servers** panel shows what is connected, what each server exposes, and which tools can write.

---

## 1. archi as an MCP server

```yaml
services:
  mcp_server:
    enabled: true
    url: "https://chat.example.org"
```

- `/mcp/sse` exposes 12 read-only tools: `archi_query` (full RAG round-trip with conversation continuity), document discovery/search/content, corpus stats, deployment info, agent specs, health.
- When chat-app auth is on, MCP clients authenticate with the standard OAuth flow they already implement: authorization-code + PKCE, discovery (RFC 8414), dynamic client registration (RFC 7591), protected-resource metadata (RFC 8707). Tokens live in the `mcp_tokens` table.
- `/mcp/auth` is an SSO-gated page to mint/rotate a token manually, with setup snippets for Claude Desktop / Claude Code / VS Code.
- Off by default (`enabled: false` registers no `/mcp/*` routes). With chat-app auth disabled the endpoint is open, matching the rest of the app.

## 2. Per-user tokens for external MCP servers

```yaml
mcp_servers:
  my_protected_server:
    transport: streamable_http
    url: https://mcp.example.org/mcp
    sso_auth: true
```

How it works for the user: log in → click **Authorize** (or get redirected once) → approve on the remote server → done. From then on every message they send builds an agent whose connection to that server carries **their** token.

Under the hood:
- The chat app brokers the OAuth flow (`/mcp/authorize`, `/mcp/callback`) with dynamic client registration and refresh, via `MCPOAuthService`.
- Tokens are stored per user per server, encrypted at rest with pgcrypto (`mcp_oauth_tokens`).
- No token yet (or expired beyond refresh)? The server is simply skipped — the agent still runs with its other tools.
- Agents holding user tokens are request-scoped (thread-local), never shared across users.
- `ssl_cert_host`/`ssl_cert_file` mount a CA bundle for TLS to internal servers (e.g. CERN CAs).

## 3. Service-account tokens for headless interfaces

Headless interfaces can never complete a browser OAuth flow. `service_auth` gives those contexts a deployment-level static Bearer token read from secrets:

```yaml
mcp_servers:
  unified:                 # interactive chat -> per-user token
    transport: streamable_http
    url: https://mcp.example.org/mcp/sse
    sso_auth: true
  unified_svc:             # headless interfaces -> service account
    transport: streamable_http
    url: https://mcp.example.org/mcp/sse
    service_auth: true
    token_secret: UNIFIED_SVC_MCP_TOKEN
```

Rules that keep this safe and predictable:
- `service_auth` servers attach **only when the request has no user identity**. Interactive chat always uses the per-user entry — so remote actions stay attributed to the real user, and the twin entries never collide on tool names.
- Secret unset → skipped gracefully. `sso_auth` + `service_auth` on one entry → per-user wins (with a warning).
- Caveat: with `allow_anonymous: true`, anonymous web users share the headless agent and would see service tools — keep anonymous off when using `service_auth`.

## Settings → MCP Servers panel

Backed by a new auth-gated `GET /api/mcp/status`; reuses the same gating code as the agent build, so the panel shows exactly what the agent would get.

- Per-server state: **Active** (with tool list), **Needs authorization** (one-click Authorize), **Headless only**, **No service token**, **Failed**.
- Summary bar with server/tool counts and a Refresh button.
- Every tool badged **read** or **write** (from the remote's `[WRITE OPERATION]` marker), with search and a "write only" filter.
- A warning when configured tools exceed the 128-tool per-request limit, since the agent drops the overflow.

**Before authorization** — one click to connect; service servers marked headless-only:

MCP Servers panel before authorization

**After authorization** — tool list with read/write badges, filter, and the >128-tool warning:

MCP Servers panel active with tools and write badges

**Dark theme:**

MCP Servers panel dark theme

## Schema

Six new tables — client side: `sso_tokens`, `mcp_oauth_clients`, `mcp_oauth_tokens`; server side: `mcp_tokens`, `mcp_auth_codes`, `mcp_registered_clients`. Defined in `init.sql` for fresh deployments and mirrored in `ConfigService` so existing volumes get them on upgrade, no manual migration. (`service_auth` needs no schema.)

## Testing

- `tests/unit/test_mcp_sse_tools.py` (9 tests) for the MCP SSE tool layer; unit suite shows no new failures vs `main`.
- Live smoke tests on a host-mode deployment:
  - **Server**: SSE protocol end-to-end — `initialize` → `tools/list` (12 tools) → `tools/call` incl. a full `archi_query` LLM round-trip; OAuth discovery/registration/metadata endpoints all answer correctly.
  - **Per-user client**: SSO login → authorize → callback stores an encrypted token → next message loads the remote's tools; missing-token and unreachable servers skip gracefully.
  - **Service client**: a loopback `service_auth` server loads tools on the headless build, is skipped on user builds, and is skipped when its secret is unset.
  - **Panel**: `/api/mcp/status` returns correct states and tool lists (incl. write flags) for both headless and user views.
  - **Upgrade**: on a pre-existing Postgres volume, `ConfigService` created all six tables at startup.

## Security review (self-review, fixed in-branch)

An adversarial pass before opening the PR found and fixed:

- **Authorization-code exfiltration**: `oauth_authorize` now validates `client_id` + `redirect_uri` against the registered client (RFC 6749 §4.1.2.1).
- **Auth bypass**: MCP bearer auth is now required whenever any chat-app auth is enabled (previously basic-auth-only deployments exposed MCP unauthenticated); OAuth endpoints only register when bearer auth is enforced.
- **Cross-user credential bleed**: agents with user tokens are request-scoped, not published on the shared pipeline.
- **Identity mismatch**: broker tokens key on the SSO subject id (`users.id`), not `preferred_username`.
- **Redirect loop**: token-store failures surface to the user; the post-login authorize chain tracks attempts per session.
- Smaller ones: auth codes consumed on failed PKCE; session ids not logged; `next` redirects restricted to relative paths; logout invalidates stored SSO tokens; silent refresh can't extend the session window; prominent warning when `BYOK_ENCRYPTION_KEY` falls back to another secret.

## Known limitations (follow-up candidates)

- **No per-server tool allow/deny list** — a remote with hundreds of tools overflows the 128-tool request limit and the agent drops the excess by list order. The panel surfaces this but can't curate it yet.
- **No write-tool gating** — tokens can carry write privileges and the agent can invoke write tools from a prompt; the panel flags them, but a confirmation gate (and read-only service accounts) are recommended.
- **Per-message rebuild cost** — with `sso_auth` servers, each authenticated message re-initializes MCP connections; a short-TTL per-user tool cache would remove most of this.
- **Shared-pipeline model overrides** still race under concurrency (pre-existing pattern; the credential side is fixed, the model/config side is not).
- **No refresh-rotation locking**; broker omits `resource`/`scope` params; remote tokens report no `expires_in`, so the broker assumes 1 hour and re-auths.
- `sso_tokens` has no consumer yet (stored for upcoming per-user Bearer use).


haozturk:
Closing this one in favor of https://github.com/archi-physics/archi/pull/605

PR #590, opened 2026-07-09
Disappearing Conversations Bug Fix

NamN8Tr:
The disappearing chats while moving between different conversations is now fixed. Now, when you refresh the page, it shows the process was interrupted. When you press esc it interrupts the process. You can run multiple conversations at the same time.


src/interfaces/chat_app/app.py
```diff
@@ -1195,6 +1195,69 @@ def _sanitize(text: str) -> str:
...
+        now = datetime.now(timezone.utc)
+        insert_tups = []
+        if not is_refresh:
+            insert_tups.append(
+                (service, conversation_id, user_sender, self._sanitize_text(user_content), '', '', user_msg_ts, model_provider, pipeline_used)
+            )
+        insert_tups.append(
+            (service, conversation_id, ARCHI_SENDER, '', '', '', now, model_provider, pipeline_used)
+        )
+        conn = psycopg2.connect(**self.pg_config)
```
AntuBattle: Is there perhaps a way we could use an existing pool to avoid connection to the db for every single insertion?

